PR #2, opened 2017-10-20
Virtual pins handlers

### LINKED ISSUE #1: Virtual Pins
Virtualization of pins, allowing the environment to be extended with external pin over SPI or I²C and making them transparent to user and libraries.

This can be implemented on multiple ways:

1 - Add an extra burden to the pinMode/digitalRead/digital by installing an 
`if (pin>nr_of_digital_pins) «call virtual pins handler here»`

2 - Extend the pin tables with a predefined set of allowed extra virtual pins, then the actual `if (port == NOT_A_PIN)` already present on the pin functions could be used to call the virtual pins handler with no extra burden to pin functions.

about the virtual pins handler, i sugest it to be **just** a function address (x3) pointing to a dumb function (just return).

this way virtual pins would have only the extra 3 function pointers and a simple return function allowing it to be implemented as a library

also all this can be conditionally compiled... if there's a way of introducing preprocessor defines to the compiler environment (-D parameter), last time i checked there was not an easy way of doing it on arduino IDE.

option 2 has the advantage of being compatible with code that checks the pin tables https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/Arduino.h#L177

I leave this to consideration and expect feedback of pros and cons before advancing.

I've done essays by extending the pin tables and they worked very well.

thanks

neu-rah:
Resolves https://github.com/arduino/ArduinoCore-avr/issues/1

defining alternative pin handling functions and respective triggers
installed into pin functions

all changes wraped by #defines

alternative handlers require compilation with VIRTUAL_PINS defined

digital pin functions triggers require both VIRTUAL_PINS and
VPINS_TRIGGER defined

eventually we can extend pin tables instead of installing the triggers

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -73,7 +73,7 @@ void yield(void);
   #define INTERNAL INTERNAL1V1
   #define INTERNAL2V56 9
   #define INTERNAL2V56_EXTCAP 13
-#else  
+#else
 #if defined(__AVR_ATmega1280__) || defined(__AVR_ATmega2560__) || defined(__AVR_ATmega1284__) || defined(__AVR_ATmega1284P__) || defined(__AVR_ATmega644__) || defined(__AVR_ATmega644A__) || defined(__AVR_ATmega644P__) || defined(__AVR_ATmega644PA__)
 #define INTERNAL1V1 2
 #define INTERNAL2V56 3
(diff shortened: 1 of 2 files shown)


neu-rah:
the pre-processor defines are only for testing purposes and provide an easy way of checking the impact by turning the feature on/off, after that we should remove them
this does not break compatibility with non-avr cores, they could implement it at its own pace and if desired.

neu-rah:
should I remove the testing #defines? they are only there for testing purposes.
However if Arduino IDE allows compiler level defines they can remain as an option for user to include/exclude the functionality.

PR #5, opened 2017-11-14
Make access to large receive buffers atomic

johnholman:
When SERIAL_RX_BUFFER is larger than 256 bytes, its head and tail
pointers are larger than 1 byte. As either pointer may be read by the
interrupt handler _rx_complete_irq() at any time, to avoid reading
inconsistent data all writes should be atomic. In addition, read
accesses to the head pointer should be atomic as this is updated
by the ISR. Read accesses to the tail pointer do not generally need
protection as this is not written by the ISR.

The necessary guards are currently missing, but are provided by this
commit using a new macro RX_BUFFER_ATOMIC to maintain code
readability. This macros expands to ATOMIC_BLOCK(ATOMIC_RESTORESTATE)
only when the receive buffer is larger than 256 bytes.

Thanks to compiler optimisation, the code generated for small
buffers is currently exactly the same as before, so there is no
impact for the normal use case.

Note that there are also atomicity issues when using a large transmit
buffer. This commit does not address those issues.

### DIFF
--- a/cores/arduino/HardwareSerial.cpp
+++ b/cores/arduino/HardwareSerial.cpp
@@ -77,12 +77,24 @@ void serialEventRun(void)
 #endif
 }
 
-// macro to guard critical sections when needed for large TX buffer sizes
+// Macro to guard critical sections for large transmit buffer sizes
+// Reads from the transmit buffer tail pointer and writes to either head or tail pointer
+// should be atomic in this case
 #if (SERIAL_TX_BUFFER_SIZE>256)
 #define TX_BUFFER_ATOMIC ATOMIC_BLOCK(ATOMIC_RESTORESTATE)
 #else
 #define TX_BUFFER_ATOMIC
 #endif
+
+// Macro to guard critical sections for large receive buffer sizes
+// Reads from the receive buffer head pointer and writes to either head or tail pointer
+// should be atomic in this case
+#if (SERIAL_RX_BUFFER_SIZE>256)
+#define RX_BUFFER_ATOMIC ATOMIC_BLOCK(ATOMIC_RESTORESTATE)
+#else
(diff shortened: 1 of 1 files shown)


johnholman:
The issues addressed so far are the same as in Robert Rozee's pull request
#3848 for the main Arduino repository with these differences:

1. guards are only placed on pointer reads when necessary, which may
improve performance for large read buffers slightly

2. an additional guard is placed on the final read buffer clear in 
HardwareSerial::end()

3. the RX_BUFFER_ATOMIC macro helps keep the code readable

johnholman:
@matthijskooijman @facchinm feedback on work so far would be welcome. 
Dealing with large write buffers might be more tricky and for a later commit.

The this pull request is similiar to @robert-rozee's  PR #3848 against the main repository 
but with these differences:

1. a new RX_BUFFER_ATOMIC macro helps keep the code readable

2. guards are only placed on pointer reads when necessary, which may
improve performance for large read buffers slightly

3. an additional guard is placed on the final read buffer clear in
HardwareSerial::end()

I don't know what the arrangements are for testing core code - is there a test suite or anything? 
As the generated code for small buffers is the same as before that shouldn't be an issue. For large
buffers I can confirm it compiles and the assembler listing suggests interrupts are being handled as
expected, but I've not actually run any tests to exercise the read functionality.

matthijskooijman:
Oh, I missed this PR, didn't have notifications enabled for this repo yet. No time now, but I'll try to have a look soon.

aweatherguy:
I think you are missing an atomic access to _tx_buffer_tail in the ::write() method. I believe the transmit UDRE ISR updates tail, and if tail is two bytes then there should be atomic access of it in the write() method.

PR #19, opened 2018-07-10
Created flexible SPI transfer functions

oclyke:
This is a small update for the core SPI library for the UNO. The concepts can be applied to other cores so that ultimately these functions could be ubiquitous.

Added:
- transferIn(void *bufIn, size_t count, uint8_t junk)
- transferOut(void *bufOut, size_t count)

transferIn allows reading data with a known output to the MOSI line without needing to pre-define the entire buffer.

transferOut preserves the content of a buffer when sending it, discarding what comes in through the MISO line. This is useful in using hardware SPI for control of addressable LEDs, OLED panels, etc...

Also had to update the surrounding directory for compatibility with Arduino IDE 1.8.5

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -144,29 +144,14 @@ nano.menu.cpu.atmega328=ATmega328P
 
 nano.menu.cpu.atmega328.upload.maximum_size=30720
 nano.menu.cpu.atmega328.upload.maximum_data_size=2048
-nano.menu.cpu.atmega328.upload.speed=115200
+nano.menu.cpu.atmega328.upload.speed=57600
 
 nano.menu.cpu.atmega328.bootloader.low_fuses=0xFF
 nano.menu.cpu.atmega328.bootloader.high_fuses=0xDA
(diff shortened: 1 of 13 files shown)


facchinm:
Hi @oclyke , thanks for the proposal! We'll surely evaluate it (and it looks promising compared to the other proposals). Just some stuff that needs to be fixed:
- https://github.com/arduino/ArduinoCore-avr/pull/19/commits/d7428ff520959a90fd2ada2e8d2d07c389976edb is reverting all the changed applied from 1.6.20, so it should be removed from this commit history
- remove the comments in the actual functions body, or add another comment saying why you commented them :slightly_smiling_face:

oclyke:
Hi @facchinm! Thanks a lot for taking a look. I went ahead and removed the commented lines and reverted [d7428ff](https://github.com/arduino/ArduinoCore-avr/commit/d7428ff520959a90fd2ada2e8d2d07c389976edbl)

PR #28, opened 2018-08-01
fix 6 year old bug in ipaddress.h

Rotzbua:
This bug causes multiple sram usage. Now only one instance should be in sram.

fix https://github.com/arduino/Arduino/issues/1007

### DIFF
--- a/cores/arduino/IPAddress.cpp
+++ b/cores/arduino/IPAddress.cpp
@@ -112,3 +112,4 @@ size_t IPAddress::printTo(Print& p) const
     return n;
 }
 
+const IPAddress INADDR_NONE(0, 0, 0, 0);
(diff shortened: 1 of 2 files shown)


matthijskooijman:
I wonder if this would be better fixed by making the declaration in the .h file `constexpr` and give it a value there. I believe that, to allow taking the address of the constant, it still needs a definition in the .cpp file as introduced by this PR, but then the definition does not need to assign a value.

facchinm:
Fixed by https://github.com/arduino/ArduinoCore-API/commit/3dbd7d7d338ff2d93c8b9dd8c833967fdb69b7c6 (as soon as AVR core will be based on chainsaw the patch will be applied here as well)

lxrobotics:
Closing this issue since initial port of Arduino [API](https://github.com/arduino/ArduinoCore-avr/commits/api) has been done and this issue is fixed with https://github.com/arduino/ArduinoCore-API/commit/3dbd7d7d338ff2d93c8b9dd8c833967fdb69b7c6 as noted by @facchinm .

Issue #25, opened 2018-07-28
Potentail corruption in attachInterrupt()


NormanDunbar:
The Arduino reference for `attachInterrupt()` states, in the syntax section, that this function should be called as per:
````
attachInterrupt(digitalPinToInterrupt(pin), ISR, mode); (recommended)
````
However, `digitalPinToInterrupt()` returns `NOT_AN_INTERRUPT` (or -1) if a pin other than 2 or 3 is passed. Within the code for `attachInterrupt()` in `WInterrupts.c`, I see this:

````
void attachInterrupt(uint8_t interruptNum, void (*userFunc)(void), int mode) {
  if(interruptNum < EXTERNAL_NUM_INTERRUPTS) {
      intFunc[interruptNum] = userFunc;
````
So, if called as recommended, and an invalid pin is passed, the -1is indeed less than `EXTERNAL_NUM_INTERRUPTS` (or 2) and so the access to the array is out of bounds and may corrupt whatever is in memory just before the array.

I suggest the following should (!) fix it:

````
 if(interruptNum < EXTERNAL_NUM_INTERRUPTS &&
    interruptNum >= 0) {
      intFunc[interruptNum] = userFunc;
````

That way, `NOT_AN_INTERRUPT` will be ignored.

**Comment by facchinm** [comment]
Hi @NormanDunbar ,
would you mind opening a PR with your modifications? Thanks!

**Comment by NormanDunbar** [comment]
Hi @facchinm ,

yes, no problem, I'll do that soon.

Cheers,
Norm.

**Comment by aweatherguy** [comment]
I'm a bit late to the party here -- was perusing these issues and noticed something about the proposed fix. I don't think a fix is required, or if so what it should be. 

Here's why: the interrupt number argument to attachInterrupt() is of type uint8_t -- unsigned. It is not possible for an unsigned value to be less than zero. If you code this:

```
attachInterupt( -1, .... )
```

The actual value passed to the function is not a signed -1, but an unsigned 255. Also, if you try to compile this:

```
void func( uint8_t x )
{
    if ( (x < 5) && (x >= 0)) { ... do something ... }
```

The compiler knows that (x >= 0) is always true because x is unsigned. You will get this warning from the compiler:

```
error: comparison is always true due to limited range of data type [-Werror=type-limits]
```

Anyway, I suggest that no fix is required because NOT_AN_INTERRUPT will get passed as 255 to attachInterrupt and that will fail the validity test as is.

[...]

PR #34, opened 2018-08-21
Fixed issue #25. Possible corruption in attachInterrupt.

NormanDunbar:
Pull Request created as requested for Issue #25:

`digitalPinToInterrupt()` may return `NOT_AN_INTERRUPT` if an invalid pin given.
`attachInterrupt()` and `detachInterrupt()` both need to check for this value or they risk
corrupting memory just before `intFunc[]`. Effectively corrupting `intFunc[-1]`.

### DIFF
--- a/cores/arduino/WInterrupts.c
+++ b/cores/arduino/WInterrupts.c
@@ -68,7 +68,7 @@ static volatile voidFuncPtr intFunc[EXTERNAL_NUM_INTERRUPTS] = {
 // volatile static voidFuncPtr twiIntFunc;
 
 void attachInterrupt(uint8_t interruptNum, void (*userFunc)(void), int mode) {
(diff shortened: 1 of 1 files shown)


aweatherguy:
I noticed an issue with this pull request...maybe no fix is necessary. See my comments in the originating issue:

[https://github.com/arduino/ArduinoCore-avr/issues/25](url)

NormanDunbar:
I think, due to the ongoing discussions in #25 that I'm going to discontinue this pull request as it is not _quite_ correct. Yet.

There's still more work to be done here on the basis of signed and unsigned 8 bit ints being passed around between `digitalPinToInterrupt()` and `attachInterrupt()`.

I'll be back! :-)

Cheers,
Norm.

PR #38, opened 2018-08-31
Treat narrowing conversion as warning, not error

PaulStoffregen:
Fix for https://github.com/arduino/Arduino/issues/7944

Before the upgrade to gcc 5.4, this was treated as a non-fatal warning (when warnings are enabled).  Now some libraries fail to compile.  This simply restores the gcc 4.8 & 4.9 behavior.


facchinm:
> I believe arduino-builder could implement some logic to detect if timestamps on library files are newer than the sketch source, or newer than other stuff (maybe the library.properties or src folder or....) the idea would be to detect when someone is editing the library files rather than just using a freshly installed copy. It's the library is being edited, maybe turn on all warnings. Doing this might improve the library quality across the entire Arduino ecosystem over time?

Wow, this is a very clever idea! One very simple first step could be activating warnings by default on libraries that are actually versioned (contain a .git folder? )

PR #40, opened 2018-09-05
Changes made to print and IPAddress

JunJieYong:
Allow print and println to have multiple arguments
Added toString in IPAddress.h


matthijskooijman:
A multi-argument print is also the subject of https://github.com/arduino/Arduino/pull/5829, which also has more discussion and its implementation is more complete. I'm also working on extending that PR with formatting, but I haven't published that yet.

PR #41, opened 2018-09-16
Added serial receive event

cabal:
As simple as possible.
It is copy - paste from Wire module (+ variable reset in constructor).
Used to receive data from modbus data wire, works well.


cores/arduino/HardwareSerial.h
```diff
@@ -134,6 +134,10 @@ class HardwareSerial : public Stream
     using Print::write; // pull in write(str) and write(buf, size) from Print
     operator bool() { return true; }
 
+    void (*user_onReceive)(void);
```
Floessie: The function pointer member shouldn't be public, or is there a reason for it?

cabal: No, there is no need for it.
As I wrote before: it is just copy/paste from Wire.
I'll change this and update PR.

PR #43, opened 2018-09-24
[variants] generic ATmega1284p support

feilipu:
**Generic ATmega1284p** support as [discussed on the Arduino Developer's Group](https://groups.google.com/a/arduino.cc/forum/#!topic/developers/tm257fZ0UEQ).

Board speeds at 16MHz, 20MHz, and 24.576MHz.
Bootloaders including optiboot and stk500v2.

**NO bootloader**, as this will depend on each specific implementations of the platform.

Atmega1284p `pins_arduino.h` designed to match UNO pin functions, and run simply through port Pin numbers (with exception of Port B SPI which is shifted by 2 bits to match UNO).


MCUdude:
Oh please no!

* The 1284P is not used on any board sold by Arduino (or Adafruit or Sparkfun). It's totally unecessary that everyone now would have four new "boards" in their boards menu when commercial board isn't available
* The pinout you provided is yet another new pinout. We already have the "[Bobuino](https://github.com/MCUdude/MightyCore/blob/master/avr/variants/bobuino/pins_arduino.h)" pinout which is accepted as the "UNO compatible" pinout. And the pinout you provided isn't really UNO compatible either, because digital pin 14 - 18 isn't the same as A0 - A5

feilipu:
> * The 1284P is not used on any board sold by Arduino (or Adafruit or Sparkfun).

This is not news to me. Please read the [referenced discussion on Arduino Developers](https://groups.google.com/a/arduino.cc/forum/#!topic/developers/tm257fZ0UEQ), my position on this issue, and perhaps a comment  there would be constructive too.

> * We already have the "[Bobuino](https://github.com/MCUdude/MightyCore/blob/master/avr/variants/bobuino/pins_arduino.h)" pinout which is accepted as the "UNO compatible" pinout.

Anyone who wants to use a Bobuino can use the Bobuino board description, bootloader, and other files. It is not the point of this PR to replace the 3rd party JSON descriptors, or provide a Bobuino solution.

Also, the Bobuino pinout is anything but standard and simple. Try laying out a generic or wire wrap board with the pin ordering:
 * 0,1,2,3,30,8,9,31 and then 4,5,6,7,10,11,12,13, etc.

perhaps the following looks easier to wire up....
 * 0,1,2,3,4,5,6,7, and then 14,15,8,9,10,11,12,13, etc.

> And the pinout you provided isn't really UNO compatible either, because digital pin 14 - 18 isn't the same as A0 - A5

Your statement is simply incorrect. It is hard to keep up, but [just 15 days ago, you committed the abstraction](https://github.com/MCUdude/MightyCore/commit/a97ef5d1d624a804a83e0bb2b351b6fa72e263f7) to ensure that references to analogue pins can matched to any hardware pin. This [abstraction was taken by Arduino three years ago](https://github.com/arduino/ArduinoCore-avr/commit/257ee3fd67f70a67b9568dfdbbe7677d06c13ff4). I've tested and used this abstraction back in 2016, and it works perfectly.

The pin layout proposd **is UNO compatible**, to the maximum extent possible, whilst keeping the physical wiring of a generic board really really simple. And, that is the point of this PR.

matthijskooijman:
I'm not really enthousiastic about including a "generic" board into the default set, since it is quite unclear what boards are supported by that, and in any case these boards are not official Arduino boards. Also, since this is a fairly simple change that does not require changes to the core, it would be trivial to turn this into a third-party core, to be installed using the boards manager with a custom URL, which seems rather easy to me.

In the discussion, this post captures my point nicely, I'm actually surprised that that post is by the same author as this PR :-) https://groups.google.com/a/arduino.cc/d/msg/developers/tm257fZ0UEQ/TE4fd1BMCgAJ

Note that I'm note a core developer, I'm just expressing my personal opinion here.

PR #48, opened 2018-11-05
Suggested improvements for micros() and the timer0 overflow ISR

aweatherguy:
This is a suggestion for speeding up micros() a bit and a slight reduction in overall system interrupt latency. 

There are a lot of comments added about the changes to help understand the intent. These may be cleaned up and/or removed as desired (if the pull request is accepted).

1) The 32-bit timer0_overflow_count has been renamed to timer0_micros and now counts in microseconds directly. The rename might be okay as no other parts of the core refer to it. However, since timer0_overflow_count was not static, it is possible that clients (i.e. sketches) are referring to it in which case the old name s/b kept.

2) With the change above, micros() no longer needs a 32-bit multiply to compute the result. Since the 32-bit counter is now directly in microseconds, all that is required is a 16-bit multiply of TCNT0 to add to the counter. This saves well over 20 instructions (call to 32-bit multiply library function) during execution, but the overall code is a few bytes larger.

3) micros() keeps interrupts disabled for a shorter length of time now, resulting in slightly improved overall
   interrupt latency for other interrupts.

### DIFF
--- a/cores/arduino/wiring.c
+++ b/cores/arduino/wiring.c
@@ -24,6 +24,7 @@
 
 // the prescaler is set so that timer0 ticks every 64 clock cycles, and the
 // the overflow handler is called every 256 ticks.
+#define MICROSECONDS_PER_TIMER0_TICK     (clockCyclesToMicroseconds(64))
 #define MICROSECONDS_PER_TIMER0_OVERFLOW (clockCyclesToMicroseconds(64 * 256))
 
 // the whole number of milliseconds per timer0 overflow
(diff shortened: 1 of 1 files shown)


aweatherguy:
Regarding (1) above, I realized that if clients (sketches) are referring to timer0_overflow, it doesn't really matter if that's renamed or not -- changing the units from overflow count to microseconds would break them anyway. The validity of this whole idea depends on whether clients are using timer0_overflow_count or not (or if they are, is it okay to break them).

PR #58, opened 2019-01-04
Support extra parameter on `attachInterrupt()`

paulo-raca:
When writing an arduino library, it is difficult to connect interrupt handlers with the component instance that should be notified.

In C, the common pattern to fix this is to specifying a `void*` parameter with the callback, where the listener can store any context necessary.

This patch adds the new function `attachInterruptParam()` to implement this pattern.

This PR was originally developed on https://github.com/arduino/Arduino/pull/4519, and it's contains lengthy discussions about the implementation strategies and overhead.

A similar PR is available for SAM devices: https://github.com/arduino/ArduinoCore-sam/pull/44

Overhead summary:
- CPU: it needs 2 extra instructions (4 cycles) before the user funcion is called.
- RAM: intFuncParam takes `2*EXTERNAL_NUM_INTERRUPTS` extra bytes (10 bytes of RAM on Arduino Pro Micro)
- Flash: Adds 56 bytes of Flash on Arduino Pro Micro

I do believe these are very acceptable values for a very useful feature.

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -148,6 +148,7 @@ void shiftOut(uint8_t dataPin, uint8_t clockPin, uint8_t bitOrder, uint8_t val);
 uint8_t shiftIn(uint8_t dataPin, uint8_t clockPin, uint8_t bitOrder);
 
 void attachInterrupt(uint8_t, void (*)(void), int mode);
+void attachInterruptParam(uint8_t, void (*)(void*), int mode, void* param);
 void detachInterrupt(uint8_t);
 
 void setup(void);
--- a/cores/arduino/WInterrupts.c
+++ b/cores/arduino/WInterrupts.c
@@ -32,10 +32,12 @@
 
 #include "wiring_private.h"
 
-static void nothing(void) {
+typedef void (*voidFuncPtrParam)(void*);
+
+static void nothing(void* arg) {
 }
 
-static volatile voidFuncPtr intFunc[EXTERNAL_NUM_INTERRUPTS] = {
+static volatile voidFuncPtrParam intFunc[EXTERNAL_NUM_INTERRUPTS] = {
 #if EXTERNAL_NUM_INTERRUPTS > 8
     #warning There are more than 8 external interrupts. Some callbacks may not be initialized.
     nothing,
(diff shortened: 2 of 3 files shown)


PaulStoffregen:
STM32 and ESP8266 are using C++ std::function to accomplish this.  I am planning to do the same on Teensy in the near future.

https://github.com/stm32duino/Arduino_Core_STM32/pull/159

https://github.com/esp8266/Arduino/blob/master/cores/esp8266/FunctionalInterrupt.cpp

paulo-raca:
Nice!
`std::function` is definitely better, unfortunately it doesn't exist on AVR, so I cannot reuse it. :(

If `attachInterruptParam` gets added, we can add a simple shim on these platforms for API consistency, something like this:

```
inline void attachInterruptParam(uint8_t pin, void (*)(void*) callback, int mode, void* param) {
    attachInterrupt(pin, [callback, param]() { callback(param); }, mode);
}
```
(I haven't really tested it)

paulo-raca:
I've just made a minor updated to this PR:
- Rebased the source against latest `master`
- Changed the comment on `attachInterrupt()`, as discusse with @aaronw@ on https://github.com/arduino/Arduino/pull/4519

Harvie:
Any news on this one?

facchinm:
Hi @Harvie ,
a slightly different version of this PR (with the same capabilities indeed) was merged in API core (https://github.com/arduino/ArduinoCore-API/blob/master/api/Interrupts.h#L19).
For the feature to rollout on every core we need to find the time for moving to API (here's the relevant branch https://github.com/arduino/ArduinoCore-avr/tree/namespaced_api)

Harvie:
@facchinm cool! can't wait for this to be available on Arduino Nano/Uno.

matthijskooijman:
I think this PR can be closed, since this will be migrated from the ArduinoCore-API repo rather than implemented separately here in any case. In the meanwhile, we have #85 and https://github.com/arduino/ArduinoCore-API/issues/95 to track this issue.

@facchinm, agreed?

matthijskooijman:
> I think this PR can be closed, since this will be migrated from the ArduinoCore-API repo rather than implemented separately here in any case. In the meanwhile, we have #85 and arduino/ArduinoCore-API#95 to track this issue.

Seems this is not really true: ArduinoCore-API has some code, but that is only for converting a reference argument to a pointer argument. And it specifies an `attachInterruptParam` function, but of course does not provide any implementation for it, which could be taken from this PR.

For additional thoughts on the API and implementation, see https://github.com/arduino/ArduinoCore-API/issues/99

paulo-raca:
It's been 5 years since I first sent this PR 😳

Time to give up

huster-songtao:
@paulo-raca You can try https://github.com/neu-rah/PCINT

PR #67, opened 2019-02-01
Changed operator bool to explicit rather than implicit conversion in Serial

pratikpc:
If you were to notice that in every Serial based Class, we have a function with the type
`operator bool() { return true; }` 

By writing explicit, we can ensure that there is no form of implicit conversion or otherwise.
For further details view https://stackoverflow.com/questions/39995573/when-can-i-use-explicit-operator-bool-without-a-cast

This would ensure that `bool b = Serial1; `would not result in b getting a value

For example, comparing the following codes:-
```
struct T {
    operator bool() const { return true; }
};
int main()
{
    T t;
    bool B = t;
}
```
Does not result in an error due to an absence of explicit.
However, adding explicit we get an error that the conversion cannot be performed in the following sample.

```
struct T {
    explicit operator bool() const { return true; }
};
int main()
{
    T t;
    bool B = t; // error cannot convert t from bool
}
```

An Operator bool with explicit would render it compatible with [only these statements](https://stackoverflow.com/a/39995574)

For example
```
while(Serial);
if(Serial)
for(;Serial;)
```

**CONSTEXPR**

Further, I have marked these statements as `constexpr `given they indeed are constant compile time expressions and their values will not change dynamically at runtime

### DIFF
--- a/cores/arduino/Client.h
+++ b/cores/arduino/Client.h
@@ -37,7 +37,7 @@ public:
   virtual void flush() = 0;
   virtual void stop() = 0;
   virtual uint8_t connected() = 0;
-  virtual operator bool() = 0;
+  virtual explicit operator bool() = 0;
 protected:
   uint8_t* rawIPAddress(IPAddress& addr) { return addr.raw_address(); };
 };
--- a/cores/arduino/HardwareSerial.h
+++ b/cores/arduino/HardwareSerial.h
@@ -132,7 +132,7 @@ class HardwareSerial : public Stream
     inline size_t write(unsigned int n) { return write((uint8_t)n); }
     inline size_t write(int n) { return write((uint8_t)n); }
     using Print::write; // pull in write(str) and write(buf, size) from Print
-    operator bool() { return true; }
+    explicit constexpr operator bool() { return true; }
 
     // Interrupt handlers - Not intended to be called externally
     inline void _rx_complete_irq(void);
--- a/cores/arduino/USBAPI.h
+++ b/cores/arduino/USBAPI.h
@@ -105,7 +105,7 @@ public:
 	virtual size_t write(uint8_t);
 	virtual size_t write(const uint8_t*, size_t);
 	using Print::write; // pull in write(str) and write(buf, size) from Print
-	operator bool();
+	explicit operator bool();
 
 	volatile uint8_t _rx_buffer_head;
 	volatile uint8_t _rx_buffer_tail;
--- a/libraries/SoftwareSerial/src/SoftwareSerial.h
+++ b/libraries/SoftwareSerial/src/SoftwareSerial.h
@@ -103,7 +103,7 @@ public:
   virtual int read();
   virtual int available();
   virtual void flush();
-  operator bool() { return true; }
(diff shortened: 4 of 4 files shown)


bxparks:
Drive-by comment:

I'm concerned that this potentially breaks user code, if someone did something like this:
```
bool status = Serial1 && (some other long boolean expression);
if (status) { ... }
```

I'm always in favor of making C++ less dangerous. But, is breaking of backwards compatibility worth it here? Is allowing `bool b = Serial1` that dangerous?

Also, it seems to me that we would need to change the `operator bool()` in the following files for consistency:
```
hardware/arduino/avr/cores/arduino/Client.h
hardware/arduino/avr/cores/arduino/USBAPI.h
hardware/arduino/avr/libraries/SoftwareSerial/src/SoftwareSerial.h
```

Additionally, how do we handle the 3rd party Cores (e.g. Teensy, ESP8266, ESP32, etc) which have implemented the `Stream` or `Serial` interface, and copied the `operator bool()` method? For consistency, don't we want to update those as well? Or are you proposing a fork in the API from those other cores?

PaulStoffregen:
> Additionally, how do we handle the 3rd party Cores (e.g. Teensy, ESP8266, ESP32, etc) which have implemented the Stream or Serial interface, and copied the operator bool() method?

Hopefully opening issues or pull requests on those other repositories isn't considered to be too much trouble?  The time to do so is when Arduino merges this change.  But don't expect any of us to actually do much at that moment.  The time to "bump" the issue with another comment is when Arduino makes a non-beta release with this feature.

I can't speak for the ESP cores, but I can tell regarding Teensy... I do pay attention to Arduino development.  I generally see changes like this long before they're merged.  :-)

pratikpc:
@bxparks actually that example will not break.
Implicit will only break direct conversion.
For example, check out the [following example on Godbolt](https://godbolt.org/z/E95xsR)

```
struct X
{
   constexpr explicit operator bool() const
   {
      return true;
   }
};

int main()
{
   X Serial;
   bool B = Serial && 34 == 34;
}
```
This example would run and work because this is not an implicit conversion but rather a form of checking a condition.

So don't worry... None of yours or the existing communities code would end up breaking!!!!
It would only break if you were doing something like
`bool B = Serial;`
This without a doubt looks fishy

I also plan on adding this Pull Request on Multiple Arduino based Repos.
I have already added a few to the official Arduino based repos after seeing Paul's recommendations to another individual.
I realise compatibility with Arduino is of pristine importance for every Arduino based fork like Teensy

pratikpc:
> The time to do so is when Arduino merges this change. But don't expect any of us to actually do much at that moment.
I understand.
I added Pull Requests to the default Arduino based repos as well after I read your comments to someone else.

pratikpc:
@bxparks in my commit, I have further updated it for 
HardwareSerial.h
SoftwareSerial.h

I was unaware Client.h and USBAPI.h also implement the function.
I will update the repo with the above correction

pratikpc:
@bxparks updated with Client.h and USBAPI.h operator bool marked as explicit as well

PR #81, opened 2019-04-06
Single pin SoftwareSerial setup

nutrun:
Allow SoftwareSerial to use only one pin (RX or TX) instead of two when a project only needs one way SoftwareSerial communication.


nutrun:
This pull request contains commits with an email from my previous job. Closing it and will create another.

PR #82, opened 2019-04-26
Minor optimization in shiftOut function

luco5826:
Hi!
I found this -very-little optimization in a function I was using with a 595 shift registers' project: while I was trying to understand what this function used to do, I figured out a (IMHO) cleaner solution.

Compiling an example with `gcc -S` generated a file 10-15 instruction lines shorter than the previous version, which could lead (i think) to a smaller workload for such a simple operation.

I didn't quite understand the double exclamation marks before the shift operation `!!(val & ...` so I don't know if this breaks something. Let me know!
Keep up with Arduino 'cause it's awesome!

### DIFF
--- a/cores/arduino/wiring_shift.c
+++ b/cores/arduino/wiring_shift.c
@@ -46,7 +46,7 @@ void shiftOut(uint8_t dataPin, uint8_t clockPin, uint8_t bitOrder, uint8_t val)
 			digitalWrite(dataPin, val & 1);
 			val >>= 1;
 		} else {	
-			digitalWrite(dataPin, val & 128);
+			digitalWrite(dataPin, (val & 128) != 0);
 			val <<= 1;
 		}
 			
--- a/cores/arduino/wiring_shift.c
+++ b/cores/arduino/wiring_shift.c
@@ -42,10 +42,13 @@ void shiftOut(uint8_t dataPin, uint8_t clockPin, uint8_t bitOrder, uint8_t val)
 	uint8_t i;
 
 	for (i = 0; i < 8; i++)  {
-		if (bitOrder == LSBFIRST)
-			digitalWrite(dataPin, !!(val & (1 << i)));
-		else	
-			digitalWrite(dataPin, !!(val & (1 << (7 - i))));
+		if (bitOrder == LSBFIRST) {
+			digitalWrite(dataPin, val & 1);
+			val >>= 1;
+		} else {	
+			digitalWrite(dataPin, val & 128);
+			val <<= 1;
+		}
 			
 		digitalWrite(clockPin, HIGH);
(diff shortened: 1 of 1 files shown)


luco5826:
Hi @matthijskooijman thank you for your review! 
I don't know why, but for a moment I tought that `val & 128` would return 1 or 0, not 10000000 or 0. Maybe we could add the `!!`  in the MSBFIRST in order to be consistent with the API, or maybe even `(val >> 8) & 1` would do the trick but then it would be very similar to the old version!
Thank you again!

bxparks:
Good clean up. But it feels slightly wrong to depend on an undocumented API behavior, to allow `(val & 128)` to act as HIGH. For one thing, this could trigger a latent bug when someone refactors digitalWrite() in the future. Secondly, Arduino code can be copied into a non-AVR Core, which may not treat 128 as HIGH. Adding the double `!!` operator or the equivalent `((val & 128) !=0)` (which ought to generate the exact same machine code if the compiler optimizes correctly), seems like the safer thing to do.

luco5826:
You're right, I'm gonna fix it and commit it again! Thank you for your fast reply!

luco5826:
Ok I fixed that, now digitalWrite will receive 1 or 0 both for LSBFIRST and MSBFIRST!

luco5826:
Hi @matthijskooijman ! Right now I checked for the compiler result, there still is 13-15 lines of difference (under plain GCC). 
I don't know about AVR compiler's optimization but I think this could be useful anyway!
Do you guys can test these changes with the AVR compiler? 
Thanks for your approvals!

luco5826:
Hi everyone!
I finally managed to check compiler's optimization under `avr-gcc` and found out that with the previous version 135 lines were generated against 118 with the second one, I know it's not that much but 17 instructions less on a basic operation is not bad!

matthijskooijman:
Hey @luco5826, thanks for testing. However, I'm a bit confused. AFAICS there's three versions involved: The original, the optimized version without `!!` and the optimized version with `!!`. Which of these did you test?

You talk about the number of lines, is this the number of lines in the dissassembly output? Note that some instructions are 2 bytes and some are 4, so it might be better to compare the compiled code size instead (or, to do a local comparison, count bytes inside the `shiftOut` function)?

luco5826:
Hi @matthijskooijman !
i tested the old version agiainst the last one I committed (the one with `!= 0`)!
The output size I tested was the one generated by the `-S` option, which stands for 
```
-S                       Compile only; do not assemble or link.
```
I don't know any other ways to test the differences, I tried to compile the code with `avr-gcc` located in `arduino/hardware/tools/avr/bin` with almost the same results as with plain `gcc` (in terms of code size)!

PR #90, opened 2019-05-27
Implemented Multiple TwoWire Slave address support

knifter:
I require my device to listen to multiple slave addresses. This can be 'hacked' by directly writing the mask to the TWAMR register. But because the TwoWire library does not store the contents of the TWDR when it enters slave-receive mode I was unable to retrieve the actual address that was written to in the onReceive() handler.

I was able to incorporate both into libraries/Wire with an example and made a pull request.
I've made two changes:
 1. support for slave-address mask setting
     begin() allows for an (optional) mask parameter, which is defaulted to 0xFF.
 2. retrieving the slave address written to from Wire
     I've stored the value of TWDR when the receiver goes into TWI_SRX mode. This value can then be 
     retreived with the Wire.getLastAddress() method.

I would prefer having this address passed as param in the onReceive callback or another option (as user:tinoest mentions on the [forum](https://forum.arduino.cc/index.php?topic=168463.0)) would be to write the slave address to the rxBuffer first. But both break backwards compatibility as it would break existing sketches.

### DIFF
--- a/libraries/Wire/examples/slave_multi_receiver/slave_receiver.ino
+++ b/libraries/Wire/examples/slave_multi_receiver/slave_receiver.ino
@@ -1,0 +1,40 @@
+// Wire Multi-Address Slave Receiver
+// by Tijs van Roon
+
+// Demonstrates use of the Wire library with multiple slave addresses
+// Receives data as an I2C/TWI slave device
+// Refer to the "Wire Master Writer" example for use with this
+
+// Created 26 May 2019
+
+// This example code is in the public domain.
+
+#include <Wire.h>
+
+void setup() {
+  uint8_t addr1 = 0x08, addr2 = 0x10; // our slave addresses
+  uint8_t addr_mask = (addr1 | addr2) ^ (addr1 & addr2);
+
+  Wire.begin((uint8_t)(addr1 | addr2), addr_mask); // join i2c bus with both addresses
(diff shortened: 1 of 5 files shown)


knifter:
This is not going to be merged, is it? Because I'm cleaning up my forks..

knifter:
I've signed the CLA several times (including last year). I've just recently discovered this commit was done with another email address. I've added those to my account and tried again and again. Sorry, can't seem to sign the CLA.

per1234:
Hi @knifter. Thanks so much for your pull request.

I apologize for the inconvenience with the CLA. I believe it will be resolved by doing this:

https://docs.github.com/en/pull-requests/committing-changes-to-your-project/troubleshooting-commits/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

Hopefully the check will automatically pass once that has been done. If not, please comment here and we will investigate.

per1234:
Aha! Please disregard my advice above. I only needed to trigger a check by the CLA bot and it is now recognized. So no additional action is required on your part for the CLA.

Chaos02:
Hi, is this still active? is there a newer version of this?
I would also like my arduino to be able to respond to two addresses (I want to emulate a fan controller I fried and do some other stuff on the same arduino)

kind regards, I hope someone can point me in the right direction!

PR #96, opened 2019-07-23
Make SoftwareSerial private methods protected to allow inheritance

gregoryng:
This pull request makes a single change to SoftwareSerial: It replaces "private" keywords to "protected" to allow users to subclass the SoftwareSerial class and modify the behavior of methods.

I have made this change locally in my environment. In my application, I need to instrument SoftwareSerial to record the timestamp of when serial bytes arrive. This means modifying the recv() method, which is currently private.  

[HardwareSerial.h](https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/HardwareSerial.h) already has its "internal" methods tagged as protected, rather than private.


gregoryng:
Closing pull request pending some additional testing in my application.

PR #97, opened 2019-07-24
Make String::concat(char) put the char directly

seanzhengw:
This pull request makes a performance enhancement to `String::concat(char)`.
The original `String::concat(char)` declares a `char buf[2]` to make a null-terminated string and calls `String::concat(const char *cstr, unsigned int length)`, and then `String::concat(const char *cstr, unsigned int length)` is going to check the parameter cstr not null and length not 0, it unnecessary and not optimized by compiler.
The `String::concat(char)` now only check `reserve()` and put the char into buffer directly.

I test this code on Arduino Mega 2560 and with compile option `-O2`, `-O3` and defualt `-Os`

    void setup()
    {
        const size_t test_size = 256;
        const size_t write_size = test_size - 1;
        String str;
        str.reserve(write_size);
        uint32_t time_start = micros();
        for (size_t i = 0; i < write_size; i++)
        {
            str.concat('0');
        }
        uint32_t time_end = micros();
        Serial.print(time_end - time_start);
        Serial.println(" us");
    }

    void loop() {}

And the result

| Code | Optimize | DATA | PROGRAM | result | 
|:--:|:--:|:--:|:--:|:--:| 
| Original | -Os | 202 | 3434 | 1732 |
| New | -Os | 202 | 3422 | 1252 |
| Original | -O2 | 202 | 3428 | 1700 |
| New | -O2 | 202 | 3428 | 1220 |
| Original | -O3 | 202 | 4826 | 1044 |
| New | -O3 | 202 | 4800 | 576 |

It shows the code size smaller and performance enhanced.

### DIFF
--- a/cores/arduino/WString.cpp
+++ b/cores/arduino/WString.cpp
@@ -279,10 +279,11 @@ unsigned char String::concat(const char *cstr)
 
 unsigned char String::concat(char c)
 {
-	char buf[2];
-	buf[0] = c;
-	buf[1] = 0;
-	return concat(buf, 1);
+	unsigned int newlen = len + 1;
+	if (!reserve(newlen)) return 0;
+	buffer[len] = c;
+	len = newlen;
+	return 1;
 }
 
 unsigned char String::concat(unsigned char num)
--- a/cores/arduino/WString.cpp
+++ b/cores/arduino/WString.cpp
@@ -282,6 +282,7 @@ unsigned char String::concat(char c)
 	unsigned int newlen = len + 1;
 	if (!reserve(newlen)) return 0;
 	buffer[len] = c;
+	buffer[newlen] = '\0';
 	len = newlen;
 	return 1;
 }


bxparks:
Good attempt, but unfortunately your code contains a small, but critical bug. The buffer inside `String` must be NUL terminated when you add a new character. So you need to add:
```C++
buffer[len+1] = '\0';
```
This will affect your benchmarks. I will guess that your new code will still be faster, but the difference may not be as much.

I'm not on the Arduino team, I don't know if they will accept your patch.

But just out of curiosity, why is the performance of `String::concat(char)` important for you? The rule of thumb is that you should avoid the `String` class if performance is critical. You will gain a lot more by rolling your code, e.g. creating a fixed sized `char[]` buffer and length parameter, adding to the buffer, then concating the whole thing to a String when you know your buffer is filled. It will be impossible to get `String::concat()` to be more than 50% the speed of rolling your own code, because `String::concat(char)` must copy 2 bytes for every 1 byte that it is asked to handle.

seanzhengw:
@bxparks thanks for pointing out my mistake, I missed the null-terminated because `String::concat(const char *cstr, unsigned int length)` use `strcpy` and my code did not.

After the fix, the new benchmarks is below.

| Code | Optimize | DATA | PROGRAM | result | 
|:--:|:--:|:--:|:--:|:--:| 
| Original | -Os | 202 | 3434 | 1732 |
| New | -Os | 202 | 3432 | 1380 |
| Original | -O2 | 202 | 3428 | 1700 |
| New | -O2 | 202 | 3438 | 1348 |
| Original | -O3 | 202 | 4826 | 1044 |
| New | -O3 | 202 | 4810 | 704 |

the code size is bigger with `-O2` but it just an option and the default is `-Os` with arduino.

The performance of `String::concat(char)` may not be important for professional programmers because they should know it's always slower than using `char[]` likes you said.
From my perspective, the Arduino Platform is a easy-to-use electronics platform, it attract many people to start programming, making something interesting, they may not be professional programmers or software/firmware engineers, the arduino core library is serving everyone, just like every programming language standard library, it's not only performance, but also improvement.

bxparks:
@seanzhengw, thanks for the new numbers. For what it's worth, I"d probably accept the PR, but the Arduino folks may prioritize simplicity over performance.

lxrobotics:
Hi @seanzhengw . First I want to thank you for creating this pull request. If you take a look into the the various `concat` implementations in `WString.cpp` you will notice that pretty every overloaded `concat` function is calling `String::concat(const char *cstr, unsigned int length)`. Arduino code does not only need to "work" but also is created under an educational point of view. Your change is creating a faster way to "concat" a char but the meaning of your implementation is not immediately clear to a user not trained in software programming. As @bxparks suggested, we are rather prioritizing simplicity over performance which is why I unfortunately have to reject your PR.

PR #284, opened 2019-09-18
Revert "Changed linking order, so precompiled libraries can be used"

lxrobotics:
Reverts arduino/ArduinoCore-avr#52 .


matthijskooijman:
For future reference: It might have been good to include a rationale for reverting in the commit messages (ideally), or at least in the PR description (but please don't force push to change the commit message now ;-p).

For anyone coming back here later, see https://github.com/arduino/ArduinoCore-avr/pull/52#issuecomment-532644517 for rationale.

PR #288, opened 2019-09-19
Made TwoWire functions virtual

### LINKED ISSUE #94: [Wire] make TwoWire all class functions virtual
To enable software I2C instances to inherit from the class without object slicing.

SRGDamia1:
Fix #94 

I'd like the functions to be virtual to allow software I2C instances to inherit from the class without object slicing.


aentinger:
Superseded by #396.

PR #297, opened 2019-11-08
added optional custom handler to Serial interrupt

gicking:
added optional custom handler to Serial interrupt. Is minor enhancement
of NeoHWSerial by SlashDevin
(https://github.com/SlashDevin/NeoHWSerial), as user function also gets
UART status as parameter to e.g. detect LIN break in user routine.

Added as core extension for consistency with Due, because AFAIK finctionality on Due cannot be implemented as library. See respective pull request for Due (https://github.com/arduino/ArduinoCore-sam/pull/93).


gicking:
replaced by new pull request with additional Serialx.attachInterrupt_Send(). Also for SAM

PR #299, opened 2019-11-23
Serial attach interrupt

gicking:
Added option to attach custom functions to Serial receive and transmit interrupts. 
Receive is similar to [NeoHWSerial by SlashDevin](https://github.com/SlashDevin/NeoHWSerial) but
  - works transparently on AVR and SAM
  - user function is called after reception of each byte
  - in addition to received byte, also UART status is passed. This allows e.g. frame synchronization via framing errors as in LIN 

Transmit is new:
  - works transparently on AVR and SAM
  - user function is called after all bytes have been send and transmission is complete
  - this allows e.g. custom protocol handlers in interrupts (like LIN)

Attached please find an example (tested for Mega and Due). Connect Tx1 & Rx1 and check serial monitor. A separate pull request was started for SAM (see [here](https://github.com/arduino/ArduinoCore-sam/pull/95)).
[Serial_attach_interrupt.zip](https://github.com/arduino/ArduinoCore-avr/files/3882493/Serial_attach_interrupt.zip)


PaulStoffregen:
Are the user functions called from interrupt context?

gicking:
yes. I need this for LIN master and LIN slave in background.

PaulStoffregen:
As a general purpose Arduino API, exposing users to more interrupt context callbacks (the API already has some), especially for data handling, is a huge mistake.

gicking:
may I ask why? Specifically how do I detect a LIN sync break if the normal receive interrupt "forgets" the status byte and only allows polling operation? SerialEvent() also doesn't help.

gicking:
actually this behaves similar to attachInterrupt(), only for Serial receive and transmit. So what's your concern exactly?

sauttefk:
I do advocate these changes, as the TX-empty interrupt has to be used for RS485 transmitter disable.

gicking:
@PaulStoffregen: actually I can see you point. Actually I started this as an extension of the [NeoHWSerial](https://github.com/SlashDevin/NeoHWSerial) **library**, i.e. not a core extension.

And for AVR obviously a library works, as NeoHWSerial shows. However, I had trouble getting it to work for the Arduino Due. If you see a way to define new "NeoSerial" instances which don't collide with the original Serial's, please let me know. I would be happy to do this as a library instead.

matthijskooijman:
> And for AVR obviously a library works, as NeoHWSerial shows. However, I had trouble getting it to work for the Arduino Due. If you see a way to define new "NeoSerial" instances which don't collide with the original Serial's, please let me know. I would be happy to do this as a library instead.

For AVR, we applied a trick defining each Serial instances along with their ISR(s), which combined well with the .a file compilation done for the core, results in the ISR only being pulled in if the corresponding serial object is actually used. However, even without that trick, if you would not use *any* Serial object, the ISRs should not be pulled in and you would be free to use a third party library for serial. Does that sound familiar? If it works when not using any serial objects, then we might be able to apply the same trick to the Due core as well to better separate the serial objects?

kcl93:
> 
> 
> As a general purpose Arduino API, exposing users to more interrupt context callbacks (the API already has some), especially for data handling, is a huge mistake.

It might be possible to leave the interrupt context on the AVR by just calling sei() before executing the user interrupt function. This way no further interrupts are going to be blocked.

I would also endorse adding an AttachInterrupt() function to the official Arduino libraries as this seems to be a commonly useful functionality to me. Relying on third party implementations for such a small change (related amount of code changed/added) will only mean increased maintaining effort.

gicking:
> > As a general purpose Arduino API, exposing users to more interrupt context callbacks (the API already has some), especially for data handling, is a huge mistake.
> 
> It might be possible to leave the interrupt context on the AVR by just calling sei() before executing the user interrupt function. This way no further interrupts are going to be blocked.
> 
> I would also endorse adding an AttachInterrupt() function to the official Arduino libraries as this seems to be a commonly useful functionality to me. Relying on third party implementations for such a small change (related amount of code changed/added) will only mean increased maintaining effort.

I agree with kcl93. The proposed core extension is very small and doesn't do anything unless the user explicitly attaches a callback function to the Serial handlers. IMHO this is a very useful - and sometimes necessary - functionality.

Implementing it as a library creates a big maintenance effort, as e.g. the SAM core still seems to be "dynamic". Also, if these became standard API functions, the developers for other, also non-Arduino boards, would also implement them over time. But if not, someone would need to create and maintain a respective library for each board, which may not happen and may even not be possible (as currently on Due).

I know that I am biased on this topic, but still I advocate to add these 2 functions to the standard Arduino-API. Whether or not this is done via merging this pull request, or implemented by a proper programmer, I don't really care. 

Thanks a lot and have a great day  :-)

**@matthijskooijman:** anyway, if whoever is in charge of deciding about this question decides against my above proposal, could you then please separate Serial the way you described? With your advice I would then create and maintain a respective library.

PS: who does decide about such an API change anyway? I assume there is some kind of "decision board" to assert a stable API...? And how long does such a decision generally take?

gicking:
**@matthijskooijman:** coming back to your comment from [Nov 24th](https://github.com/arduino/ArduinoCore-avr/pull/299#issuecomment-557921167). For me it would be acceptable (actually preferred) to separate Serial from the rest of the code, like for the AVRs. That would allow replacing the built-in UART-ISRs via a library, similar to [NeoHWSerial](https://github.com/SlashDevin/NeoHWSerial) for the AVRs. Any idea how to implement this - or maybe even a rough timeline...?

matthijskooijman:
> PS: who does decide about such an API change anyway? I assume there is some kind of "decision board" to assert a stable API...? And how long does such a decision generally take?

I do no think there is an official (documented) process for this. Discussing on the arduino-developers mailing list is usually suggested, though the core developers do not participate so actively there (though they do read it, I think). In the end, it is up to the core devs such as @cmaglie or @facchinm to decide whether or not to include something, but they have a lot on their plate already.

If this is something that should be added to all Arduino platforms (i.e. when it is a portable API, https://github.com/arduino/ArduinoCore-API/ might be the right place to discuss this (and it seems there is already a virtual HardwareSerial base class there).

> could you then please separate Serial the way you described?

I will not have the time to actually implement this, too much things on my plate already. However, if anyone else wants to take a stab at this, I'm happy to support them with advice and reviews.

To keep the discussion clean, I created new issues for the "separate serial objects" implementation here: https://github.com/arduino/ArduinoCore-sam/issues/100 and https://github.com/arduino/ArduinoCore-samd/issues/489. Let's keep *this* issue about adding the serial interrupt API.

gicking:
**@matthijskooijman**: if the serials for SAM can be implemented similar to AVR, Rx- and Tx-interrupts can be added via a library like [NeoHWSerial](https://github.com/SlashDevin/NeoHWSerial). In that case I would withdraw this pull request.

For now I propose to leave it open and discuss it in the above issues [arduino/ArduinoCore-sam#100](https://github.com/arduino/ArduinoCore-sam/issues/100) and [arduino/ArduinoCore-samd#489](https://github.com/arduino/ArduinoCore-samd/issues/489), you created.

gicking:
seriously? It takes 16 months to automatically send a legal form...???

PR #304, opened 2019-12-14
SoftwareSerial workaround for ATmega644/P

pranjal-joshi:
alt text
**ATmega644 or ATmega644P** with **Sanguino bootloader** was having compiling issues after including a SoftwareSerial library.
With this quick fix, the same library can be now compiled and use for ATmega644/P without having issues with existing compatibility with other boards.

### DIFF
--- a/libraries/SoftwareSerial/src/SoftwareSerial.h
+++ b/libraries/SoftwareSerial/src/SoftwareSerial.h
@@ -47,6 +47,33 @@ http://arduiniana.org.
 #define GCC_VERSION (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
 #endif
 
+/************************************* For ATmega644P **********************/
+
+#define digitalPinToPCMSK(p)    (((p) <= 7) ? (&PCMSK2) : (((p) <= 13) ? (&PCMSK0) : (((p) <= 21) ? (&PCMSK1) : ((uint8_t *)0))))
+#define digitalPinToPCMSKbit(p) (((p) <= 7) ? (p) : (((p) <= 13) ? ((p) - 8) : ((p) - 14)))
+ 
+#if defined(__AVR_ATmega644__) || defined(__AVR_ATmega644P__)	|| defined(__AVR_ATmega644PA)
+
+#define digitalPinToPCICR(p)    (((p) >= 0 && (p) <= 31) ? (&PCICR) : ((uint8_t *)0))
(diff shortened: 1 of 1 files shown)


matthijskooijman:
This looks like the wrong place for these macros, so I'm not sure this is the right solution. I believe normally the Arduino core or variant defines these macros.

What core and/or variant are you using?

pranjal-joshi:
I am using a typical Arduino core but added bootloader and board for ATmega644/P.
ATmega644's register structure is different than of ATmega8/168/328, so tweaking this file will make things work with 644 without breaking the core itself.
Please feel free to try this by yourself before merging PR if you are not sure about this.

Hope this helps :smile:

matthijskooijman:
> I am using a typical Arduino core but added bootloader and board for ATmega644/P.

So you added your own board to boards.txt? What variant are you using? I believe these macros should be in the variant directory, e.g. like https://github.com/arduino/ArduinoCore-avr/blob/0e7fae85bdfbab43242374c0289cdd806d13b0bc/variants/standard/pins_arduino.h#L74-L79

>  Please feel free to try this by yourself before merging PR if you are not sure about this.

I have no 644 to test with, and I'm also still unsure what settings your are using exactly.

pranjal-joshi:
I am is using a **[Sanguino](https://github.com/Lauszus/Sanguino?files=1) core**. Please take a look.
Still If it's necessary to move this code to variant directory, let me know.

matthijskooijman:
The Sanguino core already provides these macros in its variant files: https://github.com/Lauszus/Sanguino/blob/4aa3893ee97b69d46c37373a3693e43d25e7dd9d/avr/variants/sanguino/pins_arduino.h#L112-L133, so I am unsure why this does not just work in the first place.

As for working with other boards: The idea is that every board that uses the official Arduino core (such as the Sanguino boards) must have a variant directory with a `pins_arduino.h` that describes all board-specific details through, among others, these macros. This information cannot be in the main core or SoftwareSerial library, since it is not processor-specific, but board-specific (different boards with the atmega644 might choose different pin numberings), which is why variant files exist.

Hence, I'm going to close this pullrequest, since I'm now quite convinced that this is the wrong place to solve the problem you have (though I'm also unsure what the problem is, since as I mentioned, the Sanguino variant files have the macros you need). I would suggest you go back to your original problem: trying this without any chance to the Arduino or Sanguino core/board files, and if that indeed does not work, try to solve that problem (possibly asking for help on the Arduino forums, or reporting a but at the Sanguino repo if it turns out the problem is there).

If you disagree, feel free to add additional comments here.

PR #307, opened 2020-01-06
Enable C++14 support by default

federicobond:
C++14 should be fully compatible with C++11 and its support is stable since GCC 5.x.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -25,7 +25,7 @@ compiler.c.elf.flags={compiler.warning_flags} -Os -g -flto -fuse-linker-plugin -
 compiler.c.elf.cmd=avr-gcc
 compiler.S.flags=-c -g -x assembler-with-cpp -flto -MMD
 compiler.cpp.cmd=avr-g++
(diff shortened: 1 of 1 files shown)


federicobond:
We could safely go as high as C++17 with the current avr-g++ version really...

PaulStoffregen:
Will any of these give us [std::function](https://en.cppreference.com/w/cpp/utility/functional/function) on AVR?  Or is that a matter of missing C++ library support?

Some 3rd party core libraries have already started using std::function for attachInterrupt.  I'm going to move Teensy in that direction soon, and start using std::function callbacks for non-blocking APIs.  Long-term, I believe this is needed for the Arduino ecosystem.  But lack of std::function on AVR is the sticking point so far...

matthijskooijman:
> Will any of these give us std::function on AVR? Or is that a matter of missing C++ library support?

No, this is just about enabling the C++14/17 language constructs in the compiler. AFAIK, stdlibc++ is still missing for AVR.

At the risk of getting off-topic, it might be feasible to implement a lightweight `std::function`-like object for Arduino. For a customer, I once built a similar thing without dynamic allocation (so only functions/lambdas up to a fixed size could be captured), that I might be able to share later.

PaulStoffregen:
> it might be feasible to implement a lightweight std::function-like object for Arduino.

After writing, I found this repository.  Looks promising....
https://github.com/winterscar/functional-avr

PR #315, opened 2020-02-29
This much delay is okay I think

thekunalsaini:
Working

### DIFF
--- a/libraries/EEPROM/examples/eeprom_read/eeprom_read.ino
+++ b/libraries/EEPROM/examples/eeprom_read/eeprom_read.ino
@@ -52,5 +52,5 @@ void loop() {
     ++address &= EEPROM.length() - 1;
   ***/
 
-  delay(500);
+  delay(300);
 }


matthijskooijman:
Your description is a bit short. Why do you think this delay needs to be changed? Also, "is okay I think" is a bit of thin rationale, what defines "okay"?

thekunalsaini:
I m trying to say that this number represents the time in milliseconds the program has to wait until moving on to the next line of code so this much time is okay for correct execution of program.

thekunalsaini:
@matthijskooijman

matthijskooijman:
And the old value was not okay?

matthijskooijman:
Don't get me wrong - It's great that you're contributing to improve things, but to help with reviewing changes (and also, to help with understanding changes when looking at the git history later), it is important to document *why* things are changing, whether there are problems fixed, etc. Sometimes things are just a little bit nicer, cleaner, prettier without actually fixing something that was broken, but then it is good to make that explicit as well :-)

thekunalsaini:
yeah it was fine thanks I closing pull request

PR #316, opened 2020-02-29
to avoid out of bound error

thekunalsaini:


matthijskooijman:
Again, not enough details here. How can an out of bounds error occur? It looks like all array indices are anded with 0xf, which means they can never be > 15, so this should work. Also, just increasing the array size would probably not be the best fix, better to fix the out of bounds access instead).

thekunalsaini:
Yeah you are right

thekunalsaini:
Better to fix the out of bounds access instead

thekunalsaini:
I have done some updation  please see sir
Thank you

libraries/EEPROM/examples/eeprom_crc/eeprom_crc.ino
```diff
@@ -34,16 +34,22 @@ void loop() {
...
 
   for (int index = 0 ; index < EEPROM.length()  ; ++index) {
+    if(index > 16)
+    {
+     Serial.print("\nOut of Bound!");
+     break;
+    }
+     
     crc = crc_table[(crc ^ EEPROM[index]) & 0x0f] ^ (crc >> 4);
     crc = crc_table[(crc ^ (EEPROM[index] >> 4)) & 0x0f] ^ (crc >> 4);
```
matthijskooijman: @thekunalsaini, if you look closely at the code, you see that `index` is used only to index the `EEPROM` itself. The result of that is split into two 4-bit parts (guaranteed using `& 0xf`), which are then used to index `crc_table`. Since only 4 bits are used, the resulting index can never be more than 15, so there is no out of bounds access.

Do you understand what I mean? Or am I missing something here?

thekunalsaini:
but sir EEPROM can go out of bound @matthijskooijman

matthijskooijman:
I don't think so, the loop index is already limited to EEPROM length: `index < EEPROM.length()`, so that should never be out of bounds? Even more, your proposed change actually prevents reading more than 16 bytes from EEPROM, which is not actually useful.

I'm going to go ahead and close this PR for now, feel free to add further comments if needed.

PR #317, opened 2020-02-29
fix

thekunalsaini:


matthijskooijman:
This commit message is *really* lacking detail.

As for the change itself, it looks ok to remove that empty statement there (maybe it's a remnant from a time where the { } where not present). Not a very big deal, though (and it certainly does not fix any bug, at best it improves the coding style).

thekunalsaini:
But improves coding style as its better.
@matthijskooijman

PR #320, opened 2020-03-19
Only include timer0 overflow ISR and initialization when it is actually used.

matthijskooijman:
This applies the same trick we used for HardwareSerial to the timer0-related functions. With this PR, if none of `delay`, `millis` or `micros` is used, the timer0 overlow ISR is not included and enabled.

This has two advantages:
 - Sketches that want to define their own ISR are now free to do so.
 - Sketches that do not need any of these timer functions, save some memory.

Note that delayMicroseconds is not included in the above list, since it does not use timer0 but does cycle counting, so can still be used without pulling in the timer0 ISR.

This is implemented by putting the timer0 related code in a separate .c file, which is only pulled in if any of these functions are referenced (this happens because the core is linked through an .a file). To prevent the (new) `init_systick()` initialization function, which is called from `init()` in `wiring.c` to pull in the file, a weak definition of `init_systick()` is included in `wiring.c` which is replaced by the real definition from `wiring_systick.c` if any of the timer functions are used.

Without this PR, compiling the bare minimum sketch gives:

```
Sketch uses 444 bytes (1%) of program storage space. Maximum is 32256 bytes.
Global variables use 9 bytes (0%) of dynamic memory, leaving 2039 bytes for local variables. Maximum is 2048 bytes.
```

With this PR applied, the result is:

```
Sketch uses 270 bytes (0%) of program storage space. Maximum is 32256 bytes.
Global variables use 0 bytes (0%) of dynamic memory, leaving 2048 bytes for local variables. Maximum is 2048 bytes.
```

This PR is the result of some discussion in the #arduino IRC channel between @rlcamp and myself. @rlcamp implemented most of this PR, I reviewed it and made some perfectioning changes.

### DIFF
--- a/cores/arduino/wiring.c
+++ b/cores/arduino/wiring.c
@@ -24,6 +24,9 @@
 
 static void init_timers();
 
+// dummy which will be replaced by linker if any of millis(), micros(), or delay() are used
+__attribute__((__weak__)) void init_systick(void) { }
+
 void init()
 {
 	// this needs to be called before setup() or some functions won't
--- a/cores/arduino/wiring.c
+++ b/cores/arduino/wiring.c
@@ -22,228 +22,22 @@
 
 #include "wiring_private.h"
 
-// the prescaler is set so that timer0 ticks every 64 clock cycles, and the
-// the overflow handler is called every 256 ticks.
-#define MICROSECONDS_PER_TIMER0_OVERFLOW (clockCyclesToMicroseconds(64 * 256))
-
-// the whole number of milliseconds per timer0 overflow
-#define MILLIS_INC (MICROSECONDS_PER_TIMER0_OVERFLOW / 1000)
-
-// the fractional number of milliseconds per timer0 overflow. we shift right
-// by three to fit these numbers into a byte. (for the clock speeds we care
-// about - 8 and 16 MHz - this doesn't lose precision.)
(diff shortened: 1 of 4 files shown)


matthijskooijman:
@facchinm, did you guys have some script to do a bulk compile of all examples or something similar? I think it would be nice to see the impact of this change against a number of examples or real-world sketches. I wrote [a script for that](https://github.com/matthijskooijman/arduino-mass-builder) a long time ago, but that's probably out of date now (should maybe be updated to arduino-cli, but  no point in that if there is something out there already).

facchinm:
This is very clever indeed, thanks you very much for submitting it! I'll forward this to the cloud team for a mass compile against Create examples.

matthijskooijman:
Ah, ther's a cloud-based mass compile tool there, nice! Any of that open? And does it allow comparing things (like compilation failures or sketch sizes with/without a PR applied)?

matthijskooijman:
> This is very clever indeed, thanks you very much for submitting it! I'll forward this to the cloud team for a mass compile against Create examples.

@facchinm, any idea about whether this has bee ndone? Would be nice to move this change forward.

facchinm:
> @facchinm, any idea about whether this has bee ndone? Would be nice to move this change forward.

No ETA yet, sorry :disappointed:

PR #332, opened 2020-04-26
Fix off-by-one in shiftIn function

dzervas:
This bug exists in the following arduino cores as well (probably all of the arduino cores?):
- [NRF52](https://github.com/adafruit/Adafruit_nRF52_Arduino) - Opened an [MR](https://github.com/adafruit/Adafruit_nRF52_Arduino/pull/493)
- [ESP8266](https://github.com/esp8266/Arduino)
- [ESP32](https://github.com/espressif/arduino-esp32)


dzervas:
this is invalid, please ignore

PR #336, opened 2020-05-01
Improved USB ZLP logic to support long bulk transfers

jakub-vanik:
I suggest following improvement in USB zero length packet logic. The reason for this is to support bulk transfers over 64 bytes without interrupting by ZLP. This is necessary for example for USB mass storage which I am currently developing. Master might starts a read command with the length exceeding Arduino memory capacity. In that case data needs to be retrieved from storage and sent to USB in chunks. Current implementation of USB core sends ZLP when the end of chunk is aligned to 64 bytes block. This terminates the bulk transfer and causes error in host driver. I tried to fix this in a way which does not affect CDC and HID. I have tested both CDC and HID on Linux. I also checked with a logic analyzer and USB communication seems fine to me.

### DIFF
--- a/cores/arduino/USBCore.cpp
+++ b/cores/arduino/USBCore.cpp
@@ -318,14 +318,14 @@ int USB_Send(u8 ep, const void* d, int len)
 					if (ep & TRANSFER_RELEASE) {
 						sendZlp = true;
 					} else {
-						_usbZlpPending[ep] = 1;
+						_usbZlpPending[ep & 7] = 1;
 					}
 				}
 			} else if ((len == 0) && (ep & TRANSFER_RELEASE)) { // ...or if forced with TRANSFER_RELEASE
 				ReleaseTX();
-				_usbZlpPending[ep] = 0;
+				_usbZlpPending[ep & 7] = 0;
 			} else {
-				_usbZlpPending[ep] = 0;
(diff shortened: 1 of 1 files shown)


Rocketct:
Hi @jakub-vanik, thank you this enhancement, can i ask an  minimal sketch to test it?

jakub-vanik:
Hi @Rocketct, I have done some testing while preparing test sketch for you and discovered a mistake. I also realized I have to modify USB_Flush too. This is because if 2 x 64 bits are sent followed by flush there is no free bank associated with the endpoint available. No ZLP was sent in such circumstance. I have done two more commits to address these issues.

jakub-vanik:
The "minimal" test setup is quite complicated. The sketch creates a simple USB interface which sends data to host sized according to request. The counterpart is C program which uses libusb-1.0 to send request and monitor response including ZLPs. I have not discovered simpler way to demonstrate this. I have tested it on Linux. I am not sure if it works on Windows. I also added output of the test compiled with and without this enhancement. You can observe the ZLP sent after each 64 bytes when original Arduino core is used. This confuses (Linux) host mass storage driver which does not expect this. [usb_zlp_test.zip](https://github.com/arduino/ArduinoCore-avr/files/4624078/usb_zlp_test.zip)

PR #345, opened 2020-05-28
Improve reading ADC result

### LINKED ISSUE #344: analogRead - reading ADC result faster
In current core adc result is read by combining two bytes ADCH and ADCL, that way adds two unnecessary commands in assembler code. But avr-gcc compiler provides a way to access 16-bit ADC register through ADC or ADCW as long as it is defined. avr/io libraries provide those defines.
```
#ifndef __ASSEMBLER__
#define ADC     _SFR_MEM16(0x78)
#endif
#define ADCW    _SFR_MEM16(0x78)
```
It is shown below the difference in reading adc result as a combining two bytes or reading directly. As you can see there are two more command LDI and OR when combining two bytes, which are completely unnecessary.
```
x= (ADCH << 8) | ADCL; 
000004B0  LDS R24,0x0078	Load direct from data space 
000004B2  LDS R18,0x0079	Load direct from data space 
000004B4  LDI R25,0x00		Load immediate 
000004B5  OR R25,R18		Logical OR
00000086  STS 0x0123,R25	Store direct to data space 
00000088  STS 0x0122,R24	Store direct to data space 

x=ADC; 
000004B0  LDS R24,0x0078		Load direct from data space 
000004B2  LDS R25,0x0079		Load direct from data space
00000086  STS 0x0123,R25		Store direct to data space 
00000088  STS 0x0122,R24		Store direct to data space
```

I can take care of this fix, if it is needed.

Vitve4:
7.3.0-atmel3.6.1-arduino7 gcc fails to optimize separate reading from ADCL and ADCH. It produces additionally three `eor`  commands or in some cases two `mov` commands in the assembly code (see discussion #344). These commands swap register contents before store them to data area. So they are completely unnecessary. Reading ADC result with `ADC` macro fixes it and gcc generates the right code.

### DIFF
--- a/cores/arduino/wiring_analog.c
+++ b/cores/arduino/wiring_analog.c
@@ -37,7 +37,6 @@ void analogReference(uint8_t mode)
 
 int analogRead(uint8_t pin)
 {
-	uint8_t low, high;
 
 #if defined(analogPinToChannel)
 #if defined(__AVR_ATmega32U4__)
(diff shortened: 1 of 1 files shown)


matthijskooijman (approved):
Changes look good to me. As analyzed in #344, the original problem is a gcc missed optimization, but since we can't easily fix that, making this change here to fix at least this particular instance of this problem seems reasonable to me. The new code is probably even easier to read than the old code, so that's a small bonus.

One thing I'd like to see improved is the commit message. It now says "Improve", but that suggests that the reading itself is maybe somehow better or more accurate. Also, the detailed commit message has no mention of the improved generated assembly, or gcc failing to optimize this in some cases, which I think would be valuable info for anyone that comes back to this commit in the future trying to figure out why this changed :-)

I verified that this indeed improves the generated assembly using some of the examples from the linked issue.

Vitve4:
Okay, I edited the commit message to give more details.

matthijskooijman:
Message looks good, but you now only edited the PR description, not the commit message. Can you also edit the commit message (if you're unsure how, you can do this using `git commit --amend` and then pushing using `--force` (or safer, `--force-with-lease`).

Vitve4:
Oh yeah, sorry, my bad, did this with the same message as well.

matthijskooijman (approved):

PR #356, opened 2020-07-11
Added code to demonstrate use of new I2C timeout feature

paynterf:
Revised to demonstrate how to use the new I2C bus timeout feature

### DIFF
--- a/libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
+++ b/libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
@@ -39,7 +39,6 @@ void setup() {
   //demonstrates use of new anti-lockup feature
   Wire.setWireTimeout(3000, true); //timeout value in uSec, true to reset I2C bus on timeout
   wireTimeoutCount = 0;
-  Wire.clearWireTimeoutFlag(); //should be cleared on startup, but...
 
   Serial.begin(9600);
   while (!Serial); // Leonardo: wait for serial monitor
--- a/libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
+++ b/libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
@@ -36,9 +36,11 @@ uint16_t wireTimeoutCount; //capture I2C bus timeout events
 void setup() {
   Wire.begin();
 
-  //demonstrates use of new anti-lockup feature
+  #if defined(WIRE_HAS_TIMEOUT)
+  // Prevent lockups on bus problems by aborting after a timeout
   Wire.setWireTimeout(3000, true); //timeout value in uSec, true to reset I2C bus on timeout
   wireTimeoutCount = 0;
+  #endif
 
   Serial.begin(9600);
   while (!Serial); // Leonardo: wait for serial monitor
@@ -46,12 +48,14 @@ void setup() {
 }
 
 void loop() {
+  #if defined(WIRE_HAS_TIMEOUT)
   if (Wire.getWireTimeoutFlag())
   {
       wireTimeoutCount++;
       Wire.clearWireTimeoutFlag();
       Serial.print("Wire timeout detected; count now "); Serial.println(wireTimeoutCount);
   }
+  #endif
 
   int nDevices = 0;
 
--- a/libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
+++ b/libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
@@ -36,14 +36,14 @@ uint16_t wireTimeoutCount; //capture I2C bus timeout events
 void setup() {
   Wire.begin();
 
+  //demonstrates use of new anti-lockup feature
+  Wire.setWireTimeout(3000, true); //timeout value in uSec, true to reset I2C bus on timeout
+  wireTimeoutCount = 0;
+  Wire.clearWireTimeoutFlag(); //should be cleared on startup, but...
+
   Serial.begin(9600);
   while (!Serial); // Leonardo: wait for serial monitor
   Serial.println("\nI2C Scanner");
-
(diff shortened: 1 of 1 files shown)


libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
```diff
@@ -29,15 +31,29 @@
...
 void setup() {
   Wire.begin();
 
   Serial.begin(9600);
   while (!Serial); // Leonardo: wait for serial monitor
   Serial.println("\nI2C Scanner");
+
+  Wire.setWireTimeout(3000, true); //timeout value in uSec, true to reset I2C bus on timeout
+  wireTimeoutCount = 0;
+  Wire.clearWireTimeoutFlag();
```
matthijskooijman, line 45: Why not put this directly after the `Wire.begin()`? Seems more clear? Also, is clearing the timeout flag really needed here? We can rely on it being cleared on startup, I think? Same for initializing the count to 0, why not just do that in the variable declaration?

matthijskooijman (approved):
Looks good to me, I left a minor comment inline.

I think it would additionally be good to just define a timeout in all master examples, so that anyone who starts by copying an example will be safe from lockups. I would think just the timeout (with autoreset) is sufficient, no need to handle the timeout flags I think. As for the timeout, maybe 3ms is fine? Or use the higher value used by SMBus (can't remember the exact value just now)?

libraries/Wire/examples/i2c_scanner/i2c_scanner.ino
paynterf: Thanks for the comment; I moved the anti-lockup code to just under 'Wire.begin()' as you suggested.  I wanted to put the timeout flag in to expose its existence to a new user; otherwise they might never know the timeout flag existed at all or how to query it.

I could initialize the variable  in the declaration, but I have learned that assuming a variable has been initialized to zero 'somewhere else' can lead to confusing/incorrect behavior.   I like making things obvious and explicit, even at the cost of an extra line or two of code.  YMMV ;-).

matthijskooijman: >  I wanted to put the timeout flag in to expose its existence to a new user; otherwise they might never know the timeout flag existed at all or how to query it.

Yeah, that's fine (and I think this might be a good example for it), I was just suggesting not doing the timeout flag handling in *all* examples (OTOH, maybe it should be).

> I like making things obvious and explicit, even at the cost of an extra line or two of code. YMMV ;-).

It does (on this particular thing, I do like explicit in general, but also concise), but it's ok :-)

What about (not) calling `clearWireTimeoutFlag()` on startup? You've now added a comment that seems less than helpful?

paynterf: Good catch!  I removed the call to clearWireTimeoutFlag() as it is explicitly cleared in Wire.setWireTimeout()

matthijskooijman (approved):
LGTM now!

bperrybap:
I'm not sure where the put the reminder about needed a macro in the Wire library so sketches and libraries can detect the existence of the timeout API functions like setWireTimeout() but a macro needs to exist so that code can tell if the timeout api functions exist.
Further comments here: https://github.com/arduino/ArduinoCore-avr/pull/107#issuecomment-646963500

matthijskooijman:
I added the macro that @bperrybap suggested in #362. Would it be good to explicitly check for this macro in these examples too? One one hand it is pointless, since these examples are shipped with a core that has the feature, but on the other hand it does result in more portable sketches and maybe raises awareness that not all cores have this timeout feature.

bperrybap:
I think clear documentation is the most important thing.
Not sure where things like these WIRE_HAS_* macros would really get documented.
I guess that it belongs on the Wire reference pages:
https://www.arduino.cc/en/reference/wire
I'm assuming that there will be 2 new pages for setWireTimeout() and clearWireTimeout()
IMO, the Wire reference documenation is very important as it shows the official API and is a guide for all the 3rd party platform/core developers as well so any new additions to the Wire API should also include updates the Wire reference page.

BTW, just noticed that end() is not listed on the Wire reference page.... oops...
The Wire end() reference page is where WIRE_HAS_END would be documented.

For these examples, I think it would be useful to at least have a brief comment in the code that mentions that the timeout API functions are not available in all platforms or in older IDE version and perhaps even show an example of using the conditional inside the comment .
OR...
**It would be even better to provide links to the specific Wire reference pages for setWireTimeout() and clearWireTimeout() in the comment above their use in the code.**
The nice thing about having a link in the comment in the code is that a user can click on it inside the IDE and the IDE will bring up a browser so the user can seamlessly see the additional documentation.

matthijskooijman:
@paynterf, I've taken the liberty to add a commit to your PR that checks for the `WIRE_HAS_TIMEOUT` macro introduced by #362 (yet to be merged).

I thought about adding the `Wire.setWireTimeout()` (probably inside `WIRE_HAS_TIMEOUT` guards) to all examples, but I wonder if that is really useful. These examples are really short and simple and already do not do any error checking at all, so if you would enable timeouts, in these examples, in case of problems they would not lockup, but just end up looping continuously without any output, which might not necessarily be any better.

Adding such a call might be useful to let people realize that there is a timeout feature, but I wonder if it is really worth the extra complexity, especially if we want to enable timeouts by default in the future anyway.

@bperrybap, @greyltc, @cmaglie, @facchinm, any thoughts on this?

bperrybap:
@matthijskooijman Good question on if the code should set the timeout.
Ideally, that type of code could benefit from a timeout but I do understand the desire to keep it simple.
Given this code works pretty well "as is" and has for a number of years now, and that the wire library is likely going to enable timeouts by default a some point, then I would lean on keeping this code simple and not doing it.

At most maybe a mention of the timeout capability in the comments?
Or some http links in the comments to the Wire documentation page for information about the Wire API and timeouts
But I have no strong feelings on this.

PR #358, opened 2020-07-22
Add LED status capability

ScaredyCat:
Add the ability to return the status of the LEDs for CAPS_LOCK, NUM_LOCK and SCROLL_LOCK. Updates to Arduino-keyboard required for this to be useful.


ScaredyCat:
Leostick didn't like it

PR #359, opened 2020-08-30
Support to multi byte request for I2C slaves

ulysse314:
This patch adds support for multi byte request for an arduino slave.

`twi_onSlaveTransmit()` is called at least once at the beginning of the request (`TW_ST_SLA_ACK` state), and then later for each extra byte requested by the master if the buffer is empty (`TW_ST_DATA_ACK` state). If the user write all the byte needed for the request on the first call  of `twi_onSlaveTransmit()`, then this function will not be called again for the transmission.

Before `twi_onSlaveTransmit()` was only called once for `TW_ST_SLA_ACK` and never again for `TW_ST_DATA_ACK` state.

### DIFF
--- a/libraries/Wire/src/utility/twi.c
+++ b/libraries/Wire/src/utility/twi.c
@@ -627,23 +627,22 @@ ISR(TWI_vect)
       twi_txBufferIndex = 0;
       // set tx buffer length to be zero, to verify if user changes it
       twi_txBufferLength = 0;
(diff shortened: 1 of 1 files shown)


ulysse314:
Hello,

Any hope to get a review?

Thanks,

ulysse314:
I'm a real human! I just signed the Contributor License Agreement.

PR #361, opened 2020-09-17
Improve and complete implementation of new/delete

### LINKED ISSUE #47: support for notrow
http://www.cplusplus.com/reference/new/nothrow/
AFAIK the arduino core doesn't really support new/delete (it just call malloc/free): https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/new.cpp

so i guess that supporting the notrow is not feasible yet

### LINKED ISSUE #287: Operator new improvements
In https://github.com/arduino/Arduino/pull/108, some improvements to the operator new provided by Arduino were suggested (since the current forms are not standards-compliant wrt error handling IIRC). The PR itself originally suggested to remove them and include uclibc++ instead, but after some discussion, we concluded that it would be better to:

- Copy the new and delete implementations from uclibc++  (including array, placement and nothrow versions).
- Change it to remove `throw(std::bad_alloc)`.
- Define a weak `std::__throw_bad_alloc` that calls `std::terminate()`
- Define a weak `std::terminate` that calls `abort()`

I think the PR itself is not really useful directly, but the dicussion on it is. I do not recall the details right now, but wanted to at least keep record of this here.

matthijskooijman:
Since #340, the Arduino core exposes a `new` header (in addition to the previously used `new.h`), but it was incomplete, causing code that previously relied on a more complete `new` header (such as provided by uclibc++) to break. This breakage, and various considerations around improving new/delete were discussed in #287.

This PR fixes this by (almost) completing the `new` header.

This also prepares for letting `new` terminate, rather than returning null on allocation failure. The standard dictates an exception should be thrown, but that is not possible on AVR, so terminating is more similar to raising an exception that is never caught (and thus probably more standards compliant). Code that needs to (portably) handle allocation failure can (after merging this PR) already start using the nothrow version of new instead of the regular one. For compatibility reasons, the code to terminate instead of returning null is disabled yet, but can be enabled using a macro now already, and should be enabled by default in the future.

This also makes some related improvements to the `__cxa_pure_virtual` and `__cxa_deleted_virtual`.

This PR fixes #287 and fixes #47.

### DIFF
--- a/cores/arduino/abi.cpp
+++ b/cores/arduino/abi.cpp
@@ -28,14 +28,9 @@ namespace std {
 }
 
 void __cxa_pure_virtual(void) {
-  // We might want to write some diagnostics to uart in this case
-  //std::terminate();
-  abort();
+  std::terminate();
 }
 
 void __cxa_deleted_virtual(void) {
-  // We might want to write some diagnostics to uart in this case
-  //std::terminate();
-  abort();
+  std::terminate();
 }
-
--- a/cores/arduino/abi.cpp
+++ b/cores/arduino/abi.cpp
@@ -21,6 +21,12 @@
 extern "C" void __cxa_pure_virtual(void) __attribute__ ((__noreturn__));
 extern "C" void __cxa_deleted_virtual(void) __attribute__ ((__noreturn__));
 
+namespace std {
+  [[gnu::weak, noreturn]] void terminate() {
+    abort();
+  }
+}
+
 void __cxa_pure_virtual(void) {
   // We might want to write some diagnostics to uart in this case
   //std::terminate();
--- a/cores/arduino/new
+++ b/cores/arduino/new
@@ -1,5 +1,31 @@
 /*
-this header is for compatibility with standard c++ header names
-so that #include<new> works as expected
+  Copyright (c) 2014 Arduino.  All right reserved.
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
+
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+  See the GNU Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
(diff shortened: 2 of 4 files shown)


matthijskooijman:
For testing this, I've used the following sketch (manually varying the constant `s`, the `NEW_TERMINATES_ON_FAILURE` macro in the "new" header and whether or not `std::terminate` was defined in the sketch):

```C++
#include <new>

constexpr size_t s = 8000;
struct foo {
  char bar[s];
};

struct C {
  int v;
  C() : v (1234) { Serial.println("    constructor"); Serial.flush(); }
  ~C() { Serial.println("    destructor"); Serial.flush(); }
};

namespace std {
  /* void terminate() {
    Serial.println("    terminate");
    Serial.flush();
  }*/
}

void setup() {
  Serial.begin(115200);
  Serial.println("Starting");
  Serial.flush();

  char* arr;
  foo* f;
  C* c;
  
  /***** nothrow new *****/
  
  // put your setup code here, to run once:
  arr = new (std::nothrow) char[s];
  Serial.print("  arr=0x");
  Serial.println((uintptr_t)arr, HEX);
  Serial.flush();
  delete[] arr;
  
  f = new (std::nothrow) foo();
  Serial.print("  f=0x");
  Serial.println((uintptr_t)f, HEX);
  Serial.flush();
  delete f;

  c = new (std::nothrow) C();
  Serial.print("  c=0x");
  Serial.println((uintptr_t)c, HEX);
  if (c) {
    Serial.print("    c.v=");
    Serial.println(c->v);
  }
  Serial.flush();
  delete c;
  
  /***** regular new *****/
  
  arr = new char[s];
  Serial.print("  arr=0x");
  Serial.println((uintptr_t)arr, HEX);
  Serial.flush();
  delete[] arr;
  
  f = new foo;
  Serial.print("  f=0x");
  Serial.println((uintptr_t)f, HEX);
  Serial.flush();
  delete f;
  
  c = new C();
  Serial.print("  c=0x");
  Serial.println((uintptr_t)c, HEX);
  if (c) {
    Serial.print("    c.v=");
    Serial.println(c->v);
  }
  Serial.flush();
  delete c;

  /***** placement (non-allocating) new *****/
  char buf[sizeof(C)];
  
  c = new (buf) C();
  Serial.print("  c=0x");
  Serial.println((uintptr_t)c, HEX);
  if (c) {
    Serial.print("    c.v=");
    Serial.println(c->v);
  }
  Serial.flush();
  c->~C();

  Serial.println("Done");
  Serial.flush();
}

void loop() {}
```

matthijskooijman:
It would be good if we could get his reviewed and merged before 1.8.14, since it fixes a regression in 1.8.13.

Issue #353, opened 2020-07-04
setWireTimeoutUs() referenced twice in documentation for setWireTimeout()


paynterf:
The documentation for setWireTimeou() references setWireTimeoutUs() instead.

matthijskooijman:
Do you mean the comments in the source code? Or is there some reference documentation online already?

That function was renamed just before merging #107, probably the documentation wasn't updated. Should be fairly trivial to fix.

paynterf:
Yep - just the comments in the Wire.cpp source code, directly above the function itself

PR #362, opened 2020-09-17
Wire: Define WIRE_HAS_TIMEOUT

matthijskooijman:
In commit deea929 (Introduce non compulsory Wire timeout), some
timeout-related functions were added. To allow writing portable
sketches, it is important for those sketch to know whether they are
using a Wire library version that is new enough to offer these functions
without having to rely on version number checking (since other Arduino
cores have their own versioning schemes).

This adds a WIRE_HAS_TIMEOUT macro, similar to the existing WIRE_HAS_END
macro, to facilitate that.

This relates to #42.

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -97,7 +97,7 @@ void TwoWire::setClock(uint32_t clock)
  * master that has claimed the bus).
  *
  * When a timeout is triggered, a flag is set that can be queried with `getWireTimeoutFlag()` and is cleared
- * when `clearWireTimeoutFlag()` or `setWireTimeoutUs()` is called.
+ * when `clearWireTimeoutFlag()` or `setWireTimeout()` is called.
  *
  * Note that this timeout can also trigger while waiting for clock stretching or waiting for a second master
  * to complete its transaction. So make sure to adapt the timeout to accomodate for those cases if needed.
@@ -105,7 +105,7 @@ void TwoWire::setClock(uint32_t clock)
  * but (much) shorter values will usually also work.
  *
  * In the future, a timeout will be enabled by default, so if you require the timeout to be disabled, it is
- * recommended you disable it by default using `setWireTimeoutUs(0)`, even though that is currently
+ * recommended you disable it by default using `setWireTimeout(0)`, even though that is currently
  * the default.
  *
  * @param timeout a timeout value in microseconds, if zero then timeout checking is disabled
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -111,7 +111,6 @@ void TwoWire::setClock(uint32_t clock)
  * @param timeout a timeout value in microseconds, if zero then timeout checking is disabled
  * @param reset_on_timeout if true then TWI interface will be automatically reset on timeout
  *                           if false then TWI interface will not be reset on timeout
-
  */
 void TwoWire::setWireTimeout(uint32_t timeout, bool reset_on_timeout){
   twi_setTimeoutInMicros(timeout, reset_on_timeout);
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -109,12 +109,12 @@ void TwoWire::setClock(uint32_t clock)
  * the default.
  *
  * @param timeout a timeout value in microseconds, if zero then timeout checking is disabled
- * @param reset_with_timeout if true then TWI interface will be automatically reset on timeout
+ * @param reset_on_timeout if true then TWI interface will be automatically reset on timeout
  *                           if false then TWI interface will not be reset on timeout
 
  */
-void TwoWire::setWireTimeout(uint32_t timeout, bool reset_with_timeout){
-  twi_setTimeoutInMicros(timeout, reset_with_timeout);
+void TwoWire::setWireTimeout(uint32_t timeout, bool reset_on_timeout){
+  twi_setTimeoutInMicros(timeout, reset_on_timeout);
 }
 
 /***
--- a/libraries/Wire/src/Wire.h
+++ b/libraries/Wire/src/Wire.h
@@ -25,6 +25,7 @@
 
 #include <inttypes.h>
 #include "Stream.h"
+#include "utility/twi.h"
 
 #define BUFFER_LENGTH 32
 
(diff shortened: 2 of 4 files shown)


matthijskooijman:
I added one more commit that also defines the default timeout in Wire.h (without changing it).

matthijskooijman:
I added one more commit that renames a parameter to be a bit more clear, I noticed this when working on a documentation proposal.

facchinm (approved):

bperrybap:
Guys, something to think about is how to handle defaults in the Wire.cpp code.
In the PR code I saw for the update to make enabling the timeout the default, it is done in begin() by explicitly setting the defaults.
This can be problematic in that begin() can and often does get called more than once.
What that means is that if one library attempts to change the default timeout settings, another library can clobber it when it calls begin()
There isn't a clear cut answer to whether begin() should force defaults, as there can be arguments made for or against it.
However, IMO, it shouldn't especially since multiple libraries can and do call it.
Not doing it in begin() means that there would be a default and that any overriding of the default always takes affect until there is another override. No call to begin() could accidentally erase an explicit override of the defaults.

The chipKit platform handles this by maintaining a reference counter that gets bumped when begin() is called and decremented when end() is called and only does special initialization the first time begin() is called.
They have some additional needs (like some h/w locks up if attempt to do it more than once)

For this platform,  If it is desirable to only set the defaults once,
I think a simple solution would be to move the call to set the timeout:
`  twi_setTimeoutInMicros(WIRE_DEFAULT_TIMEOUT, WIRE_DEFAULT_RESET_WITH_TIMEOUT);`

 out of begin() and into the constructor.

That will allow it to only set the defaults once so that subsequent calls to begin() do not clobber it.

matthijskooijman:
Good point. I also agree with your proposed solution. Even though it is not always advisable to do initialization in the constructor of a global object, in this case, the initialization is really only just setting variables, so it should be safe.

This does raise the question: Should maybe twi.c just have default values for these timeout variables instead, referring to these constants? That would be more efficient, though currently I don't think twi.c refers to Wire.h, so that would break the abstraction a bit (even though that distinction is not really useful atm anymore, but still).

bperrybap:
`This does raise the question: Should maybe twi.c just have default values for these timeout variables instead, referring to these constants? That would be more efficient, though currently I don't think twi.c refers to Wire.h, so that would break the abstraction a bit (even though that distinction is not really useful atm anymore, but still).`

Good question.

I assume that long term, that some sort of timeout support will be enabled by default so some of this is kind of more related to the transition to that point and for those of us that are worrying about backward compatibility issues.


I guess it is basically a judgement call as to where the defaults should live.
They could live in Wire or in twi. 
I.e. perhaps the defaults used in Wire could come from twi.h and and also be used to compile time initialize the timeout variables in twi.c
or Wire could be in charge and use its values regardless of any defaults in twi.c

From an API and documentation perspective, it may make sense if Wire declares and documents its own defaults and uses them.
This ensures that the Wire layer and API is always is in control of these parameters and gets what it wants/needs.
This can be important if these new timeout API functions are intended to be promoted to other platforms.
It also may be important to keep it at the Wire object level since at some point or on some platforms there may be multiple i2c interfaces and each could have its own configuration tracked in its Wire object.

I don't have very strong feelings about it.


One thing I just noticed in Wire.h and the twi.c code is that it appears that are 3 different sets of defaults being used that are different.
- #defines Wire.h
- defaults for the setWireTimeout() call
- hard coded values that compile time assigned to variables in twi.c
```
// When not configured, these settings are used for the timeout
#define WIRE_DEFAULT_TIMEOUT 0
#define WIRE_DEFAULT_RESET_WITH_TIMEOUT false
```

And then this in the TwoWire class:

`    void setWireTimeout(uint32_t timeout = 25000, bool reset_with_timeout = false);
`
Shouldn't the defaults for setWireTimeout() be same as the defines?
i.e. use the defines vs hardcoded values.
```
And in twi.c
It uses its own set of hard coded values for defaults.
```
static volatile uint32_t twi_timeout_us = 0ul;
static volatile bool twi_timed_out_flag = false;  // a timeout has been seen
static volatile bool twi_do_reset_on_timeout = false;  // reset the TWI registers on timeout
```

matthijskooijman:
> It also may be important to keep it at the Wire object level since at some point or on some platforms there may be multiple i2c interfaces and each could have its own configuration tracked in its Wire object.

Excellent point, though the code is currently not quite structured to really support this. I guess if the AVR core ever moves to this, maybe the  Wire/twi distinction should be just removed anyway (are there any AVR chips with multiple I2C interfaces at all, though?)

> From an API and documentation perspective, it may make sense if Wire declares and documents its own defaults and uses them.

Agreed, all of utility/twi is an implementation detail and should be covered by APIs in the Wire "namespace".


> Shouldn't the defaults for setWireTimeout() be same as the defines?
i.e. use the defines vs hardcoded values.

No, since the current defaults are *no* timeout, and the defaults for `setWireTimeout()` are intended to be way to say "enable a timeout with a reasonable value". If you would use the same defaults for both, then (until a timeout is enabled by default), running `setWireTimeout()` would actually disable the timeout, which is not what (AFAICS) is the intention here.


In any case, here's what I'm thinking to do: Let twi.h expose constants for the default values, which are then used in twi.c. Then Wire.h can redefine its own constants (this will be the publically usable API, and named like the constants in the current version of this PR), which are just initialized to the TWI constants, so the values match. Then, the call to `twi_setTimeoutInMicros` can be removed from begin, and I think all is well:
 - If `setWireTimeout()` is not called, the default values `WIRE_DEFAULT_TIMEOUT` and `WIRE_DEFAULT_RESET_WITH_TIMEOUT` are used.
 - If `setWireTimeout()` is called, then the last call decides the settings that are used, regardless of any subsequent calls to begin().
 - If `setWireTimeout()` is called without arguments, some reasonable timeout setting is enabled. It is not possible for sketches to query which settings these are, but if a sketch really cares, it should just pass values instead, I think.

I'll go ahead and implement the above right away, let's see how that looks.

matthijskooijman:
I just did a force push implementing the above strategy. Only commit d739b3f6f7c7c9e7f0c2dc11cc91057fef164b3d was changed, the others are unchanged (but everything is rebased on top of current master).

As far as I'm concerned, this addresses the excellent concerns raised by @bperrybap and makes this PR ready to merge again.

matthijskooijman:
I've done a force push to add a commit message reference to #353 (which is fixed by the last commit).

Can we get around to merging this? I think having a reliable way to detect feature support is an important part of (this) API, and really should not be left unfixed...

PR #364, opened 2020-09-21
Remove explicit true/false check

arnavb:
This PR should probably have been made 2 years ago, but this is a migration of arduino/Arduino#7036.


matthijskooijman:
Change looks good to me.

However, given recent developments, I think this PR would need to go into https://github.com/arduino/ArduinoCore-API (where the AVR core and all other cores will start pulling from at some point in the future).

PR #368, opened 2020-10-30
Change Nano parameters

ArminJo:
see Issue #308


matthijskooijman:
Thanks for the quick PR! However, @per1234 clarified the problem with this approach in the issue, so I'm afraid we can't really merge this...

ArminJo:
@matthijskooijman You are right

Issue #179, opened 2014-04-01
caterina bootloader


fricklr:
problem with bootloader handover:
http://www.avrfreaks.net/index.php?name=PNphpBB2&file=viewtopic&t=123862
+++ bootloaders/caterina/Caterina.c
@@ -151,9 +151,31 @@
        LEDPulse();
    }
-  // Disable the USB interface
-   USBCON &= ~(1 << USBE);
-   
-   // Disable the VBUS transition enable bit
-   USBCON &= ~(1 << VBUSTE);
-   
-   // Disable the VUSB pad
-   USBCON &= ~(1 << OTGPADE);
-   
-   // Freeze the USB clock
-   USBCON &= ~(1 << FRZCLK);
-   
-   // Disable USB pad regulator
-   UHWCON &= ~(1 << UVREGE);
-   
-   // Clear the IVBUS Transition Interrupt flag
-   USBINT &= ~(1 << VBUSTI);
-   
  +

cmaglie:
I've looked at the linked forum thread, and it seems that the issue was fixed in LUFA at the times, so I'm closing this one. Feel free to comment further if I'm wrong.

teliot:
Ran into this issue when I disabled USB which made the MCU unresponsive once sei() was called as USB interrupts were triggering but not correctly handled.

Changing the Caterina boot loader would fix the issue for others in the future. I have no idea if this would require a matching change in the Arduino USB CDC code to correctly re-enable the USB endpoints.

156a157,161
> 	/* clear leftover USB settings from bootloader or application interrupts will fail with USB disabled*/
> 	USBCON = 0;
> 	USBINT &= ~(1 << VBUSTI);
> 	UHWCON &= ~(1 << UVREGE);
>

PR #383, opened 2021-01-06
Allow disabling CDC with -DCDC_DISABLED

DanielGibson:
This change makes it relatively easy to disable the CDC at compiletime.
Disabling it of course means that the serial console won't work anymore, and that you can only flash directly after resetting the board, but I think that's not much of a problem.

`CDC_DISABLED` is also used in ArduinoCore-samd for the same purpose,
see https://github.com/arduino/ArduinoCore-samd/blob/master/cores/arduino/USB/USBDesc.h#L24-L27

Disabling CDC helps making Arduino-based USB devices work with old or otherwise problematic hardware or software.
For example, USB Boot Keyboards don't work with many KVM switches if the CDC sub-devices are there; I'm sure there's other systems with the same problem (BIOS/UEFI implementations? Old operating systems - someone mentioned assistive devices still running Windows XP). Apart from Keyboards, I read that disabling CDC for Arduino-based MIDI devices.

based on https://github.com/gdsports/usb-metamorph/tree/master/USBSerPassThruLine

See also https://github.com/NicoHood/HID/issues/225 and https://github.com/arduino/Arduino/issues/6387 and https://forum.arduino.cc/index.php?topic=545288.msg3717028#msg3717028

### DIFF
--- a/cores/arduino/CDC.cpp
+++ b/cores/arduino/CDC.cpp
@@ -22,6 +22,13 @@
 
 #if defined(USBCON)
 
+#ifndef CDC_ENABLED
+
+#warning "! Disabled serial console via USB (CDC)!"
+#warning "! With this change you'll have to use the Arduino's reset button/pin to flash (upload)!"
+
+#else // CDC not disabled
+
 typedef struct
 {
 	u32	dwDTERate;
@@ -299,4 +306,5 @@ int32_t Serial_::readBreak() {
 
 Serial_ Serial;
 
+#endif /* if defined(CDC_ENABLED) */
 #endif /* if defined(USBCON) */
--- a/cores/arduino/USBCore.cpp
+++ b/cores/arduino/USBCore.cpp
@@ -69,8 +69,18 @@ const u8 STRING_MANUFACTURER[] PROGMEM = USB_MANUFACTURER;
 #define DEVICE_CLASS 0x02
 
 //	DEVICE DESCRIPTOR
+
+#ifdef CDC_ENABLED
 const DeviceDescriptor USB_DeviceDescriptorIAD =
 	D_DEVICE(0xEF,0x02,0x01,64,USB_VID,USB_PID,0x100,IMANUFACTURER,IPRODUCT,ISERIAL,1);
+#else // CDC_DISABLED
+// The default descriptor uses USB class OxEF, subclass 0x02 with protocol 1
+// which means "Interface Association Descriptor" - that's needed for the CDC,
+// but doesn't make much sense as a default for custom devices when CDC is disabled.
+// (0x00 means "Use class information in the Interface Descriptors" which should be generally ok)
+const DeviceDescriptor USB_DeviceDescriptorIAD =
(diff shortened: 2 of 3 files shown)


teliot:
This rolls back facchinm@a1ba49e7c8d1067d3f0a87d54aedd2a746fe3342

DanielGibson:
Why was this feature removed in the first place?

It's very useful, after all.

teliot:
It looks what was removed was a define to enable/disable the CDC endpoint. My assumption is the current USB capabilities were made to work awaiting further improvement...

I can see that around the same period a few changes were made such as PluggableUSB and HID which forces a CDC endpoint. This is restricting and would be nice to be able to expand USB capabilities without blowing up the size of the compiled binaries. However its been like this for 5 years now.

DanielGibson:
> I can see that around the same period a few changes were made such as PluggableUSB and HID which forces a CDC endpoint. 

I'm not sure I understand - with my patch I'm using HID without CDC being enabled?

mdevaev:
@facchinm Could you take a look at this? The problem really exists and affects several projects that are working on boot hid devices.

NicoHood:
I 2nd this PR. It would help a lot of people reporting issue with the HID Keyboard at bios time!

SukkoPera:
+1

DanielGibson:
Any chance this can get merged?
If not, what should I change?

pinging @facchinm @silvanocerza

teliot:
Found an issue with the change.
The caterina bootloader is not removing the USB interrupts. Without code clearing the USB interrupts once sei() is called my sparkfun pro micro unit locks up. Should clear these interrupts if CDC_DISABLED is defined.
#179 
USBCON = 0;
USBINT &= ~(1 << VBUSTI);
UHWCON &= ~(1 << UVREGE);

DanielGibson:
Interesting - I didn't have that problem with my cheap Pro Micro clone - do you get it from just disabling CDC or only if you completeley disable USB?

DanielGibson:
Also, what changes exactly must be done?
https://github.com/arduino/ArduinoCore-avr/blob/master/bootloaders/caterina/Caterina.c doesn't really look like anything mentioned in #179 (or your comment): USBCON, USBINT and UHWCON aren't used at all

teliot:
I was having issues when i removed the USB code and was using hardware serial. The USB interrupts were still firing away and locked up at the first sei(); until i changed those registers. I will do a proper test now using your CDC_DISABLED patch but i suspect the same result as caterina bootloader is leaving usb in a semi initialized state.

teliot:
I tested the patch and found no issues as it does still leave USB connected to the host which is different then what I have been doing.

SukkoPera:
I've also been using this patch for a while and it really works like a charm, please merge.

facchinm:
Merged, sorry for the incredibly long delay

rapapar:
Any of all solution does not work with pico zero, it's why I find a way:
go to 
Arduino15\packages\rp2040\hardware\rp2040\3.7.0\libraries\Adafruit_TinyUSB_Arduino\src\arduino\Adafruit_USBD_CDC.cpp
search first uint16_t Adafruit_USBD_CDC::getInterfaceDescriptor(
add a definitive return 0; at the 1sty line, 
byebye CDC

PR #391, opened 2021-01-23
Update math macros to avoid computing callable arguments more than once

Keating950:
Currently, if you pass an a callable argument to (e.g.) the `max` macro, the argument will be computed more than once. `max(expensive_function(a), expensive_function(b))` will expand to:

```c
expensive_function(a) > expensive_function(b) ? expensive_function(a) : expensive_function(b)
```

The compiler can't always optimize these calls away, as seen in the [Godbolt disassembly here](https://godbolt.org/z/WMbbdd). This PR uses the GCC [`typeof`](https://gcc.gnu.org/onlinedocs/gcc/Typeof.html) extension to avoid these unnecessary calls where possible, falling back to the existing macros should that feature be unavailable. (Testing for it with the latest Arduino SDK shows that the extension is available.)

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -89,14 +89,40 @@ void yield(void);
 #undef abs
 #endif
 
-#define min(a,b) ((a)<(b)?(a):(b))
-#define max(a,b) ((a)>(b)?(a):(b))
-#define abs(x) ((x)>0?(x):-(x))
-#define constrain(amt,low,high) ((amt)<(low)?(low):((amt)>(high)?(high):(amt)))
-#define round(x)     ((x)>=0?(long)((x)+0.5):(long)((x)-0.5))
+#ifdef __GNUC__
+	#define max(a,b) \
+	({ typeof (a) _a = (a); \
+		typeof (b) _b = (b); \
+		_a > _b ? _a : _b; })
+	#define min(a,b) \
+	({ typeof (a) _a = (a); \
+		typeof (b) _b = (b); \
+		_a < _b ? _a : _b; })
+	#define abs(a) \
+	({ typeof (a) _a = (a); \
+		_a > 0 ? _a : _a * -1; })
+	#define constrain(amt, low, high) \
+	({ typeof (amt) _amt = (amt); \
+		typeof (low) _low = (low); \
+		typeof (high) _high = (high); \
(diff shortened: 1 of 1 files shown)


cores/arduino/Arduino.h
```diff
@@ -89,14 +89,40 @@ void yield(void);
...
+	#define min(a,b) ((a)<(b)?(a):(b))
+	#define max(a,b) ((a)>(b)?(a):(b))
+	#define abs(x) ((x)>0?(x):-(x))
+	#define constrain(amt,low,high) ((amt)<(low)?(low):((amt)>(high)?(high):(amt)))
+	#define round(x)     ((x)>=0?(long)((x)+0.5):(long)((x)-0.5))
+	#define sq(x) ((x)*(x))
+#endif
+
 #define radians(deg) ((deg)*DEG_TO_RAD)
 #define degrees(rad) ((rad)*RAD_TO_DEG)
```
maaler, line 125: These two lines (124, 125 for radians and degrees) expose the very same problem as do *all* macro calls. Why not define them as inline constexpr functions instead? Like so: 
'
`template<typename T>
inline constexpr 
T min(T a, T b) {
    return a > b ? b : a;
}`

matthijskooijman: Because that doesn't work in C, and these macros are currently usable from C, so that would break compatibility.

matthijskooijman:
Note that min/max have already been updated similarly to what you do in this PR (and with a template function for C++) in the 
https://github.com/arduino/ArduinoCore-API repository. That repo will be used as a basis for the common code in this repo in the (near?) future, so that would be the best place to make fixes like these (I don't think changes to the common code will be merged here anymore).

As I said, just `min` and `max` have been fixed in that repo, the others are still up for improvement. See https://github.com/arduino/ArduinoCore-API/issues/85 for discussion about this, and a partial implementation that could be used as a basis. @Keating950, if you're interested in this, maybe you could make a PR over there? The -API repo is already used by the megaavr and SAMD boards, so it's easy to test there, and I think there's also a branch in this repo that used the -API repo (but it might be a bit old).

cores/arduino/Arduino.h
Keating950: I was also concerned about binary size.

Keating950:
That sounds like the best place to start. Thanks for the direction!

PR #397, opened 2021-03-05
Finish USBDevice.detach() placeholder.

teliot:
Only have Arduino Micro board to test on.

The ide fails to make the attempt to upload a sketch if the COM port is unavailable after detaching the USB serial CDC endpoint.

The RunnerException error in SerialUploader.java:152 should be caught and handled when attempting to upload to a nonexistent COM port. A console output should be given instructing the user to press the reset button, power cycle, or plug the board into a USB port so that the bootloader can upload a sketch from the IDE. This would allow a developer to upload sketches even when the sketch disables USB, or takes full control of the keyboard or mouse preventing the user from using their computer.

For example; if USBDevice.detach() is called in setup() a user would need to instruct avrdude on the command line and then get the device into the bootloader to upload a sketch. In practice the CDC endpoint resets the device and the IDE waits for the bootloader to upload the code. The bootloader can be physically triggered and a sketch uploaded if the IDE ignores the error when attempting to open the serial port and send the reset signal.

### DIFF
--- a/cores/arduino/USBAPI.h
+++ b/cores/arduino/USBAPI.h
@@ -62,6 +62,7 @@ public:
 	bool configured();
 
 	void attach();
+	// IDE CANNOT UPLOAD OVER USB WHILE DETACHED
 	void detach();	// Serial port goes down too...
 	void poll();
 	bool wakeupHost(); // returns false, when wakeup cannot be processed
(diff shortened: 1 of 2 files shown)


Dakkaron:
Is there any reason this has not been merged yet? I was stuck on USBDevice.detach() not working for like 10h until I found this.

PR #400, opened 2021-03-22
Make EEPROM variable 'extern'

### LINKED ISSUE #399: EEPROM variable declared static instead of extern
```warning: 'EEPROM' defined but not used [-Wunused-variable]```

This warning occurs because EEPROM.h declares the EEPROM variable as follows:

```static EEPROMClass EEPROM;```

This means that the EEPROM variable is local (not exported) from every file that includes EEPROM.h.

I'm [developing a program](https://github.com/jamessynge/tiny-alpaca-server) using the Arduino framework that has lots of source files. I do unit testing on a linux host, so I find it convenient to have a platform.h header file that includes either the Arduino framework header files that are used by the source file, or my minimal host versions of those Arduino files for execution on host. Among those files is EEPROM.h, so for every .cpp files that includes my platform.h file but doesn't actually use EEPROM I see the warning above. The fix is quite simple, and I'm happy to provide a PR if the team is interested.

jamessynge:
Fixes https://github.com/arduino/ArduinoCore-avr/issues/399, a compiler warning in any compilation unit where EEPROM.h is directly or indirectly included by not used.

### DIFF
--- a/libraries/EEPROM/src/EEPROM.cpp
+++ b/libraries/EEPROM/src/EEPROM.cpp
@@ -1,0 +1,4 @@
+#include "EEPROM.h"
+
+EEPROMClass EEPROM;
+
--- a/libraries/EEPROM/src/EEPROM.h
+++ b/libraries/EEPROM/src/EEPROM.h
@@ -142,5 +142,5 @@ struct EEPROMClass{
     }
 };
 
-static EEPROMClass EEPROM;
+extern EEPROMClass EEPROM;
 #endif


sauttefk:
+1

matthijskooijman:
This change looks good at first glance.

However, `EEPROMClass` really is a class with just methods (so no members that require storage, though I think an object needs at least one byte anyway), so I wonder if the use of `static` here might not help the compiler to optimize away the `EEPROM` object (since it knows that the `static` object cannot be used in any other compilation units, so can be tossed if not needed). OTOH, LTO might make this a moot point, or adding `static` might actually do the reverse if the compiler cannot optimize away the object, then using EEPROM from multiple compilation units will result in multiple duplicate EEPROM objects.

Could you maybe test this? I think a good test would be a sketch that uses EEPROM from two different compilation units (i.e. two .cpp files, two .ino files are joined into one compilation unit) and then compile it with and without this PR applied. If the compilation with this PR applied used less or the same amount of memory, then I'm all for merging this.

matthijskooijman:
Also: If the memory usage does become worse, an alternative fix for this could be the gcc `unused` attribute to suppress the warning (with a good comment explaining why it is needed).

jamessynge:
OK, I've tested as requested and found no difference in the amount of Flash or RAM. You can find the test cases in [this commit](https://github.com/jamessynge/arduino_experiments/commit/d6967f71817b29680764f1930919622f0806f914).

jamessynge:
Just woke up in the middle of the night with the thought that I made a mistake in that test. Hold off until I verify tomorrow.

jamessynge:
I did make a mistake, but having updated the test programs and ensured that I was modifying the correct copy of the EEPROM.h file, the conclusion is the same: **there is no change in the size of the program or the RAM usage, whether the variable is extern or static.**

I'm a bit perplexed why having `static EEPROMclass EEPROM;` in multiple compilation units doesn't result in an increase in RAM size, but perhaps the compiler is determining that the code never uses the address of the EEPROM variable. I'm not at all sure if that is a thing the compiler can do.

matthijskooijman (approved):
> I'm a bit perplexed why having static EEPROMclass EEPROM; in multiple compilation units doesn't result in an increase in RAM size, but perhaps the compiler is determining that the code never uses the address of the EEPROM variable. I'm not at all sure if that is a thing the compiler can do.

Yeah, I think that's probably what happens indeed.

Thanks for the thorough testing, looks like you covered any reservations I could have made, so I think this looks good to merge. It's not up to me to merge this, though, but I'll mark my approval and leave it to others to pick this up for merging :-)

PR #420, opened 2021-07-19
Update Stream.cpp to handle bytes in range 0x80 to 0xff

drmpf:
Fixes reading bytes in the range 0x80 to 0xff

### DIFF
--- a/cores/arduino/Stream.cpp
+++ b/cores/arduino/Stream.cpp
@@ -33,8 +33,10 @@ int Stream::timedRead()
   int c;
   _startMillis = millis();
   do {
-    c = read();
-    if (c >= 0) return c;
+  	if (available()) { // skip read of -1's
+      c = read();
+      return c & (0x00ff); // prevent sign extension from 0xff return
+    }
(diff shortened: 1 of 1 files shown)


yumkam:
I don't see why this is necessary/correct.
`read()/peek()` is supposed to return either char as non-negative integer in range of 0 to 255 (e.g. cast from `unsigned char` or `uint8_t` to `int`), or `-1` in case of error/timeout. `< 0` check is correct.
If some derived class cast from (signed) `char` to `int` (and return chars in `[\x80-\xff]` range as negative int), then it is broken and must be fixed; this PR only (partially) hides the problem (and moves problem up call-chain).
(As far as I can see, `HardwareSerial`, `SoftwareSerial`, `CDC` and `Wire` are correct).
(That said, I noticed `Stream::parseInt` and `Stream::parseFloat` are broken - they mishandle `ignore = '\xff'` [and it is NOT fixed by this PR either]).

drmpf:
You are correct. I 'fixed' a problem that does not exist.  Closing this.

PR #422, opened 2021-07-23
HID bugfix: SET_IDLE uses wValueH for duration, not wValueL

mdevaev:
The duration in the SET_IDLE request is actually responsible for the highest byte, not the lowest. Here is an excerpt from the standard:
image

image

To verify this, I used [USBCV](https://www.usb.org/document-library/usb3cvx64-tool), the official tool for testing compliance with USB standards. Without this patch, the "Idle Test" was a failure, everything works as it should with the patch.), the official tool for testing compliance with USB standards. Without this patch, the "Idle Test" was a failure, everything works as it should with the patch.

### DIFF
--- a/libraries/HID/src/HID.cpp
+++ b/libraries/HID/src/HID.cpp
@@ -128,7 +128,7 @@ bool HID_::setup(USBSetup& setup)
 			return true;
 		}
 		if (request == HID_SET_IDLE) {
-			idle = setup.wValueL;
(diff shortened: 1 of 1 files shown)


NicoHood:
It is very likely that the ARM core and all other cores need that fix as well!

mdevaev:
@per1234 Could you take a look at this? This is a one-line fix.

PR #425, opened 2021-08-09
Add docs folder (SoftwareSerial)

canchebagur:


MakerPact:
@per1234 This is showing up as still open. Do you need help with this? if so what, and where should I look into it at?

per1234:
Hi @MaderDash. No we don't need help with this.

If we need help with a pull request, we will add a statement to that effect. So if you don't see such a statement, please assume no help is needed.

You are welcome to look through the issues in our repositories to find opportunities to contribute. You should generally be self-directed in such contributions. If you have to ask on GitHub for assistance to make a contribution then it is likely the project maintainers would end up putting more effort into assisting you than we would just doing it ourselves. Of course there is value in mentorship, but we simply don't have any resources to dedicate to that at this time.

You are always welcome to request assistance or open discussions over on Arduino Forum:

https://forum.arduino.cc/

PR #441, opened 2021-12-02
Add another Mega2560 Board, with CP2012

softhack007:
This patch adds the PID/VID for my "KS0499 Mega 2560 Plus Board" (keyestudio). 
It is a mega2560,  with has the CP2012 as USB interface. For more Information, see  https://wiki.keyestudio.com/KS0499_KEYESTUDIO_Mega_2560_Plus_Board


WestfW:
Won't that cause all boards using the cp2102 (which I assume is what you meant?) to identify as a MEGA?
It's a generic USB/Serial converter - is keystudio modifying the vid/pid explicitly to indentify their board (if that's even possible)?
It's not necessary for the IDE to identify a board with a generic USB/Serial adapter specifically, in order to use it.  It will just show up in the PORTS list as an unidentified serial port.

softhack007:
:thinking: actually I don't know if keyestudio created a unique vid/pid or not ... but for sure it would be stupid if all boards with this genetic usb converter would identify as MEGA. 

Your the experts :monocle_face: - if my vid/pid is not uniquely identifying the board model, I would agree that my PR gets closed without being merged into the core.

PaulStoffregen:
According to this site, 10C4:EA60 is the ID for all Silicon Labs CP210x USB-to-UART chips.

https://devicehunt.com/view/type/usb/vendor/10C4/device/EA60

PR #442, opened 2021-12-17
Add drivers originally bundled with Java IDE directly in the core

umbynos:
Originally the drivers for windows were bundled in the [Java IDE](https://github.com/arduino/Arduino/tree/master/build/windows/dist/drivers), historically the avr core was the one bundled with the IDE, which was responsible for installing them. With the new IDE 2.0 the avr core is downloaded and installed during the first execution.
It makes much more sense to leverage the [`post-install.bat` ](https://arduino.github.io/arduino-cli/dev/platform-specification/#post-install-script) script to do this task, freeing the IDE from this responsibility.
During the process of porting the drivers I removed CP210x drivers since we do not have anymore boards that require them, and also the `Old_Arduino_Drivers.zip` since they were unused.
I also reorganized a bit the folders: now all the gemma drivers and libusb (required by the gemma) are under `gemma/` folder.

To test my changes I created a fake index located [here](https://gist.githubusercontent.com/umbynos/38dc547e3a90020c86ea3b8f95af4955/raw/77d2633a6abfaf795f2cd16e1e1a9bce9eb0b171/package_index_drivers_avr.json)
and installed it using the java IDE on a PC with win 10.
I added the `allow_insecure_packages=true` option in the `preferences.txt` to allow the `post-install.bat` script to be run and since the fake index does not have signature. The result form my testing are:
 
immagine
immagine
Before:
immagine

After: 
immagine


umbynos:
immagine
It works with the mini uno too

cmaglie (approved):

PR #459, opened 2022-02-20
Software serial half duplex parity

AwonSawsay:


per1234 (changes requested):
Introduces a significant breaking change to the API, regressions, and pointless formatting changes.

You should also provide more information about the proposal. It appears to be a copy/paste from an existing variant of the library, but you don't mention the source.

You need to take more care with your proposals if they are to be considered @AwonSawsay

PR #467, opened 2022-03-09
add Serial rx callback function

lshw:
example:

```cpp
//return false drop char,
bool rxcap(char ch) {
  if (ch == '1') { //if get char '1'
    Serial.println("callback!");
    Serial.setRxCallBack(0); //disable CallBack
    return false; //del char '1' from rx buff
  }
  return true;
}

void setup() {
  Serial.begin(115200);
  Serial.setRxCallBack(rxcap);
}

void loop() {
while(Serial.available())
  Serial.write(Serial.read());
}
```
~

### DIFF
--- a/cores/arduino/HardwareSerial.h
+++ b/cores/arduino/HardwareSerial.h
@@ -106,6 +106,7 @@ class HardwareSerial : public Stream
     volatile rx_buffer_index_t _rx_buffer_tail;
     volatile tx_buffer_index_t _tx_buffer_head;
     volatile tx_buffer_index_t _tx_buffer_tail;
+    bool (*_rxcallback)(char);
 
     // Don't put any members after these buffers, since only the first
     // 32 bytes of this struct can be accessed quickly using the ldd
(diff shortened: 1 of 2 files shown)


lshw:
This function can perform real-time operations, such as xon/xoff,
or When avr uses multiple serial ports, use this function to realize that each serial port has a different size of receive buffer,
or real-time mirroring from one serial port to another serial port,
or can filter the received content first. refill the system's receive buffer,

matthijskooijman:
See also https://github.com/arduino/ArduinoCore-API/issues/160 for some discussion on adding such a callback to the API more broadly.

PaulStoffregen:
This example code calls Serial.print() from within the callback, which unintentionally illustrates the main reason why callbacks directly from interrupts are not a good design for the Arduino API.

I'm not against callback.  In fact, I want to see this API.  But it really should be done in another way where the interrupt merely sets a flag or other mechanism for the callback to happen later, from main program context (eg, made from yield or an RTOS mechanism).

PR #473, opened 2022-04-27
Improve F macro to avoid string duplication in Flash. Fix #472.

### LINKED ISSUE #472: Improve F macro to avoid string duplication in Flash
Based on [this](http://michael-buschbeck.github.io/arduino/2013/10/20/string-merging-pstr/) blog post, one can improve the `F` so it avoid string duplication in Flash.

Orace:
See related links:
 http://michael-buschbeck.github.io/arduino/2013/10/20/string-merging-pstr/
 http://michael-buschbeck.github.io/arduino/2013/10/22/string-merging-pstr-percent-codes/

### DIFF
--- a/cores/arduino/WString.h
+++ b/cores/arduino/WString.h
@@ -35,7 +35,23 @@
 //     -std=c++0x
 
 class __FlashStringHelper;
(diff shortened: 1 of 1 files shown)


Orace:
@michael-buschbeck should be set as author.

Orace:
Maybe it's PSTR that should be improved

PR #480, opened 2022-06-15
SoftwareSerial: Add support for half-duplex

### LINKED ISSUE #205: Allow SoftwareSerial to share RX and TX pins. [imported]
This is [Issue 1022](http://code.google.com/p/arduino/issues/detail?id=1022) moved from a Google Code project.
Added by 2012-08-30T14:00:13.000Z by [dmel...@gmail.com](http://code.google.com/u/107726014984682000895/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Enhancement, Priority-Medium, Component-SoftwareSerial
### Original description

**What change would like to see?**

From Ed Baafi. The ability for SoftwareSerial to use the same pin for RX and TX (half-duplex). See: https://github.com/Modkit/Arduino/commits/new-extension for a possible implementation.

It would be good if this worked well / simply if you only wanted to either transmit or receive, but not both.

**Why?**

To reduce pin usage.

**Would this cause any incompatibilities with previous versions?  If so, how**
**can these be mitigated?**

dominikberse:
resolves #205

This pull request allows using the same pin for RX/TX in half-duplex mode. By using `listen` and `stopListening` one can switch between reading and writing. Calling `write` while listening will instantly return 0 (currently without setting a write error).
 
I tested bidirectional communication using a Mega2560 and an Uno connected via a single wire.

### DIFF
--- a/libraries/SoftwareSerial/src/SoftwareSerial.cpp
+++ b/libraries/SoftwareSerial/src/SoftwareSerial.cpp
@@ -439,7 +439,7 @@ int SoftwareSerial::available()
 size_t SoftwareSerial::write(uint8_t b)
 {
   if (_singleWirePin && isListening())
-    setupTXPin();
+    setupTXPin(_singleWirePin);
 
   if (_tx_delay == 0) {
     setWriteError();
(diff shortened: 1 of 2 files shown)


dominikberse:
It seems that the Arduino SoftwareSerial libraries for STM32/LPC1768 behave a little different. The serial is in TX mode by default, calling `listen()` will automatically put it into RX mode after each transmission. I adapted to this behavior, since for my use case (Marlin) a consistent behavior is better. (Calling `listen()` without send will not put the serial in RX mode, see https://github.com/stm32duino/Arduino_Core_STM32/pull/1771)

PR #482, opened 2022-06-17
ESP core `printf` implementation in AVR core.

jfjlaros:
This addition will allow for convenient printing using the `printf` function, as implemented in the ESP core.

**Example**

Consider the following block of code.
```cpp
Serial.print("(");
Serial.print(i);
Serial.print(", ");
Serial.print(j);
Serial.print("): ");
Serial.println(result);
```

This can be replaced by the following line.

```cpp
Serial.printf("(%i, %i): %i\r\n", i, j, result);
```

I hope you find this useful.

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -264,3 +264,33 @@ size_t Print::printFloat(double number, uint8_t digits)
   
   return n;
 }
+
+size_t Print::printf(const char *format, ...)
+{
+    char loc_buf[64];
+    char * temp = loc_buf;
+    va_list arg;
+    va_list copy;
+    va_start(arg, format);
+    va_copy(copy, arg);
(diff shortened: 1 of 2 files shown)


matthijskooijman:
Thanks for this PR. However, in the past, adding printf support has been discussed and rejected, see https://github.com/arduino/ArduinoCore-API/issues/32 which tracks this issue and has some additional references, and also talks about better approaches.

In any case - if this would be implemented it would need to be supported on all architectures and its implementation would be arch-independent anyway, so the proper place to add (and/or discuss) this would be the ArduinoCore-API repository (which is not used by AVR yet, but should be in the future). In fact, that repo already has some PRs for implementing this (including one that is technically a bit cleaner than this PR, because it prints byte-by-byte without requiring a fixed-size buffer).

I'm going to go ahead and close this PR for these reasons, but know that your contributions are very much appreciated!

PR #493, opened 2022-07-26
The patches required to use C++ STL

KOLANICH:
These patches should be enough to use `magic_enum` library and `std::string`. For using CTRE other patches are needed, but they don't belong to this repo.


KOLANICH:
It is likely these patches should be applied to other cores too.

KOLANICH:
CLA won't be signed, but everyone can consider original parts of my contribution as licensed under Unlicense license.

matthijskooijman:
Thanks for your contribution!

I believe this might be already solved in https://github.com/arduino/ArduinoCore-API, which should be the base of this core as well in the future (also see https://github.com/arduino/ArduinoCore-API/pull/140 which makes some additional related changes).

KOLANICH:
Thanks. Incorporated some changes from there.

PR #494, opened 2022-07-26
Replace macros with more modern constructs

KOLANICH:
Macros are big issues because they are not "AST-safe" and wreck the things. This set of commits replaces them with more modern and more safe mostly (for sane use-cases) backward-compatible analogues in order to reduce the probability of unintended effects happening.

It is likely this set of commits should be applied to other cores too.

CLA won't be signed (because it is cancer destroying free software), everyone can consider any original contributions within this commits as licensed under Unlicense license.


matthijskooijman:
Part of these changes are already made in https://github.com/arduino/ArduinoCore-API, though not all of them. I believe that if these changes are made, they should be made in that repo (and this repo should start using that one, hopefully soon...).

One consideration is that existing code might be using `#ifdef(PI)` or so, which would break by your changes. This could be worked around by doing something like `const PI = 3.14; #define PI PI`.

PR #514, opened 2022-12-06
Make ISR(Timer0_OVF_vect) a weak function

drshihan:
In order to let people use Timer0 with all its functions, mark this with the attribute 'weak', so they can redefine this function with their own code.


matthijskooijman:
There is already another PR with another approach that should have the same result here: #320. Personally I think that that approach is slightly cleaner (and also more in line with how the HardwareSerial ISRs work). Maybe you could test that PR and report there whether it also fixes your issue?

PR #528, opened 2023-03-18
use less stack on Print::printNumber

DEvil0000:
The issue with printNumber is that it uses a lot of stack quite unexpectedly. This makes it basically unusable on a attiny.
I suggest doing the naive implementation as it's done with float print. 

I think the Autor was hoping for a more optimized write call Implementation in the over-writable write call with a long buffer. I think this is too late to get it more efficient for most platforms anyway. Another option of course would be changing the API and adding a parameter for this or adding a define to switch the default Implementation.

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -208,7 +208,7 @@ size_t Print::printNumber(unsigned long n, uint8_t base)
   }
 
   // prevent crash if called with base == 1
-  if (base < 2) base = 10;
+  if (base < 2) return printNumber(n, 10);
 
 // use -D ARDUINO_PRINT_NUMBER_GENERIC_ONLY when compiling to get only the generic version
 #ifndef ARDUINO_PRINT_NUMBER_GENERIC_ONLY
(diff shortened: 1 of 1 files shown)


per1234:
Hi @DEvil0000. Thanks for your pull request. I'm just commenting to let you know the failure of the "Spell Check" workflow run is not related to the changes you proposed (it was caused by a recent automated update of the spellcheck dictionary) and won't affect the review of the PR.

I have now fixed the problem in the `master` branch: https://github.com/arduino/ArduinoCore-avr/pull/529

PaulStoffregen:
This PR causes digits to print in reverse order.

Tested with Arduino Uno using IDE 1.8.19 with AVR core 1.8.6.

screenshot

screenshot2

DEvil0000:
@PaulStoffregen sorry, looks like I was a bit sleepy on the first PR. I fixed the implementation with the first two commits and added improved versions for base 2 and base 16 which one can disable by compiler define flag.

TBAMax:
Yeah, catering for all the bases with one function messes it up there. When printing decimal only, much smaller buffer is needed(11bytes vs 33bytes).
The good formula for calculating needed digits for decimal:
`#define MAX_DECIMAL_SIZE(x)  ((size_t)(8 * sizeof(x) * 302 / 1000) + 1)`

PR #530, opened 2023-03-20
WString: change strcpy() to memcpy()

mar0x:
It is worth to use memcpy() to avoid extra checks. Furthermore, it fixes possible issues when copied WString has zero byte stored intentionally.


bilaliscarioth:
bump

PR #536, opened 2023-06-05
PR: Add some macros related to working with bytes and bits

DRSDavidSoft:
### Description

This PR adds some useful macros for manipulating bytes and bits in C/C++. The existing code already had some macros for low and high bytes, bit reading, setting, clearing, toggling and writing. I added the following macros:

- `packToInt(h,l)`: This macro takes two bytes (high and low) and packs them into a 16-bit integer.
- `packToLong(i3,i2,i1,i0)`: This macro takes four bytes and packs them into a 32-bit long integer.
- `bitIsSet(value, bit)`: This macro returns true if the given bit in the value is set (1).
- `bitIsClear(value, bit)`: This macro returns true if the given bit in the value is clear (0).

These macros can be useful for working with binary data, such as serial communication, encryption, compression, etc. They can also make the code more readable and concise.

I hope you find this PR useful and I welcome any feedback or suggestions. 

<sup>Note: This excerpt was generated by ChatGPT. The code is written by me.</sup>

### DIFF
--- a/cores/arduino/Arduino.h
+++ b/cores/arduino/Arduino.h
@@ -73,7 +73,7 @@ void yield(void);
   #define INTERNAL INTERNAL1V1
   #define INTERNAL2V56 9
   #define INTERNAL2V56_EXTCAP 13
-#else  
+#else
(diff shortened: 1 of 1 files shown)


Randyblo7:
Nice

PR #542, opened 2023-09-09
Fix compilation warnings about unused parameter "tag" in file `new.cpp`

Ivan-Stefanov-513:
Fix compilation warning on file `cores/arduino/new.cpp` about unused argument `tag` in four of the functions by casting the argument to void.

### DIFF
--- a/cores/arduino/new.cpp
+++ b/cores/arduino/new.cpp
@@ -57,6 +57,7 @@ void * operator new[](std::size_t size) {
 }
 
 void * operator new(std::size_t size, const std::nothrow_t tag) noexcept {
+  (void)tag; // unused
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Hi @Ivan-Stefanov-513. Thanks for taking the time to submit this pull request.

I see we have a previous proposal for this change at https://github.com/arduino/ArduinoCore-avr/pull/474. There are also some alternative proposals:

- https://github.com/arduino/ArduinoCore-avr/pull/476
- https://github.com/arduino/ArduinoCore-avr/pull/484

It is best to have only a single PR for each change so I'll go ahead and close this in favor of the other.

PR #546, opened 2023-10-11
Reserve the correct (smaller) amount of flash for Optiboot on the Arduino Nano.

mjoldfield:
Two variants of the Arduino Nano board are supported: one using an old
boot loader; the other the new, smaller, optiboot.

However, the flags which protect memory and the memory reserved for
the boot loader when linking haven't been updated for optiboot. This
PR fixes that.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -213,12 +213,12 @@ nano.build.variant=eightanaloginputs
 ## --------------------------
 nano.menu.cpu.atmega328=ATmega328P
 
-nano.menu.cpu.atmega328.upload.maximum_size=30720
+nano.menu.cpu.atmega328.upload.maximum_size=32256
 nano.menu.cpu.atmega328.upload.maximum_data_size=2048
 nano.menu.cpu.atmega328.upload.speed=115200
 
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Thanks for taking the time to submit a pull request @mjoldfield.

We cannot accept this change because the fuses on the official Nano boards are configured for a 2 kB boot section instead of 0.5 kB.

So the current `upload.maximum_size=30720` value is actually correct even though the Optiboot bootloader used on the boards would fit in a 0.5 kB boot section.

Related discussion:

- https://github.com/arduino/Arduino/commit/1cf34c8250658c5c4a5c488f7a7a0d7071e3430f#commitcomment-27651798
- https://github.com/arduino/ArduinoCore-avr/issues/10
- https://github.com/arduino/ArduinoCore-avr/issues/308
- https://github.com/arduino/Arduino/issues/7341
- https://github.com/arduino/Arduino/issues/7482
- https://github.com/arduino/ArduinoCore-avr/pull/26
- https://github.com/arduino/Arduino/issues/8310
- https://github.com/arduino/ArduinoCore-avr/issues/98

PR #564, opened 2024-06-30
Update new.cpp

jmparatte:
Suppress 4 annoying warnings in new.cpp.

### DIFF
--- a/cores/arduino/new.cpp
+++ b/cores/arduino/new.cpp
@@ -57,6 +57,7 @@ void * operator new[](std::size_t size) {
 }
 
 void * operator new(std::size_t size, const std::nothrow_t tag) noexcept {
+  (void)tag; // unused
 #if defined(NEW_TERMINATES_ON_FAILURE)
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Thanks for taking the time to submit this fix @jmparatte. I see we already have a pull request for this exact change (https://github.com/arduino/ArduinoCore-avr/pull/474) so I'll close this as a duplicate.

matthijskooijman:
>  I think a new compilation option declared in [Preferences...]

Do you realize that there is an option in the preferences already to enable compiler warnings? It is not enabled by default, which would be good (but people were hesitant to do so, since that will also produce warnings for other people's code (libraries and cores) that a beginner user cannot really fix.

jmparatte:
> that a beginner user cannot really fix.

True. It's why I have posted 4 changes in AVR new.cpp to correct the code and suppress warnings of 4 unused parameters. Simple. But warn beginners or possible miscoding is good. It isn't?

PR #565, opened 2024-09-03
Quick Fix of overflow for larger toggle_count values

lookfwd:
Quick Fix of overflow for larger `toggle_count` values. Because of the 16-bit `int` value in many platforms including Uno, the calculation of `toggle_count` overflows easily. For example with a frequency of 39kHz and a duration of 9ms, you get `toggle_count == 112 `, which is 1.5ms.


matthijskooijman (approved):
Looks like a good fix to me.

PR #582, opened 2024-12-22
HardwareSerial: Define no-op serialEvent() functions for clang

beakthoven:


matthijskooijman:
Looks ok to me. Did you verify this still works as expected on gcc (with and without a user serialEvent detected)?

dakkshesh07:
> Looks ok to me. Did you verify this still works as expected on gcc (with and without a user serialEvent detected)?

The patch was successfully tested with GCC in Arduino ide, both with and without a user-defined serialEvent. The functionality works as expected in both cases.

Test code:

```
void setup() {
  Serial.begin(9600);
  Serial.println("Testing serialEvent()");
}

void loop() {
  if (Serial.available()) {
    Serial.println("Processing data in serialEvent()");
  }
  delay(1000);
}

void serialEvent() {
  while (Serial.available()) {
    char c = Serial.read();
    Serial.print("Received: ");
    Serial.println(c);
  }
}
```

The code compiled perfectly and i was able to test it out with Serial monitor.

A slight size reduction was observed when compiling code without a user-defined serialEvent function. This occurs because GCC now identifies the weak symbols as explicitly defined no-ops, allowing for more aggressive dead code elimination.

Issue #590, opened 2025-01-16
Replace use of strcpy by strlcpy on String::concat(const char*,uint32_t) ?


bilaliscarioth:
Hello Arduino Core Teams,

Can we replace the use of `strcpy`(2) from `String::concat(const char*, uint32_t)` by `strlcpy(2)` or even `memcpy` ?

Without referencing  `strcpy`(2) on your documentation, if we cast a `struct` into `char*`, we could had
a null terminator between the start and the end, and `strcpy`(2) will never copy at the last char.

```
unsigned char String::concat(const char *cstr, unsigned int length)
{
	unsigned int newlen = len + length;
	if (!cstr) return 0;
	if (length == 0) return 1;
	if (!reserve(newlen)) return 0;
	strcpy(buffer + len, cstr);
	len = newlen;
	return 1;
}
```

Is it by choice, by using `strcpy` ?

bilaliscarioth:
Duplicated of #530

PR #591, opened 2025-01-16
Replace strcpy by memcpy on String::concat(const char*, uint32_t);

bilaliscarioth:
That PR come from with the issues #590 , where I explain in details why strcpy is not safe.


bilaliscarioth:
Duplicated of #530...

PR #612, opened 2025-11-22
Wire: Fix write() returning incorrect byte count on buffer overflow (See issue #597)

### LINKED ISSUE #597: write() doesn't return actually sent bytes in case of buffer overflow
The buffer used for transmission is fixed at 32 bytes. Trying to use write() to transmit more than this in one transmission results in only 32 bytes being sent out, but the `write()` function incorrectly returning the amount of bytes originally requested.

Example of writing a page to a 24C EEPROM chip:
```C
void writePage(int addr, char* buf, int len) {
  Wire.beginTransmission(I2C_ADDRESS);
  Wire.write(addr);  // set write offset address
  Wire.endTransmission();
  for(int i = 0; i < len;) {
    Wire.beginTransmission(I2C_ADDRESS);
    i += Wire.write(buf + i, len - i);  // i will immediately be set to len and loop will terminate prematurely
    Wire.endTransmission();
  }
}

writePage(0, data, 256);
```
In the above example I am trying to overcome the 32 byte buffer limitation by paging writes to whatever size `write()` actually writes out, however it will always return the full length, making the code believe everything was sent out. Instead I would have to limit my transmission to `BUFFER_LENGTH` manually. `write()` should check if the requested amount of data is exceeding the buffer and then return the amount that is actually still able to be sent out.

Tranomial:
Wire.write() previously returned the number of bytes requested, even if the
internal 32-byte transmission buffer could not store all of them. This caused
code that paged or chunked writes to prematurely terminate, and silently drop
data.

The function now returns the actual number of bytes appended to the TX buffer,
matching the standard Arduino Wire API behavior and allowing callers to detect
buffer limits correctly.

Fixes #597.

### DIFF
--- a/.github/workflows/check-arduino.yml
+++ b/.github/workflows/check-arduino.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Arduino Lint
         uses: arduino/arduino-lint-action@v2
--- a/.github/workflows/compile-platform-examples.yml
+++ b/.github/workflows/compile-platform-examples.yml
@@ -214,7 +214,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Compile examples
         uses: arduino/compile-sketches@v1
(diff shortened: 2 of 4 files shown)


Tranomial:
@per1234 
Just pinging in case this PR got buried below other PRs

PR #616, opened 2026-02-17
Arduino AVR Boards (Debug enabled)

felias-fogg:
This PR contains the necessary changes to platform.txt, boards.txt, and programmer.txt to enable debugging for all MCUs (with a debugging interface) supported by the "Arduino AVR Boards" platform. The GDB server PyAvrOCD, a patched version of AVR-GDB, and a copy of simavr are hosted on my GitHub account as release assets of PyAvrOCD. If you want to take a "test drive", you have to extend the list of additional board manager URLs in the Preferences dialog by

`https://felias-fogg.github.io/ArduinoCore-avr/package_felias-fogg_ArduinoCore-avr_index.json`

and then install the `Arduino AVR Boards (Debug enabled)` platform. Documentation can be found here: [https://pyavrocd.io](https://pyavrocd.io). You can use the simavr Simulator by chosing simavr as a programmer and then start debugging. However, it is more fun to use a real debugger (one of the supported Microchip debuggers or an Arduino UNO with the dw-link sketch).

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -1,6 +1,8 @@
 # See: https://arduino.github.io/arduino-cli/latest/platform-specification/
 
 menu.cpu=Processor
+menu.jtag=JTAG pins
+menu.lto=Compiler LTO
 
 ##############################################################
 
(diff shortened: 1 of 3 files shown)


felias-fogg:
I have removed the LTO option (that I had introduced) from the tool section and moved the disabling of LTO to the "Optimize for Debugging" option. It will make things easier for the user and even with LTO enabled, debugging is still possible.

PR #621, opened 2026-04-23
enhancement: more print format constants defined; function printNumber(..) adapted.

TobiasKnauss:
- HEX2, HEX4, HEX8
- BIN2, BIN4, BIN8, BIN16, BIN32

The number represents the minimum digit count for printing, i.e. adding leading zeros.

This enhancement was inspired by https://forum.arduino.cc/t/serial-print-value-hex/463868/5

It can be tested with the code example from https://docs.arduino.cc/language-reference/en/functions/communication/serial/print/ enhanced like this:

```
/*
  Uses a for loop to print numbers in various formats.
*/
void setup()
{
  Serial.begin(9600); // open the serial port at 9600 bps:
}

void loop()
{
  // print labels
  Serial.print("NO FORMAT"); // prints a label 
  Serial.print("\t"); // prints a tab
  Serial.print("DEC"); 
  Serial.print("\t");
  Serial.print("HEX"); 
  Serial.print("\t");
  Serial.print("HEX2"); 
  Serial.print("\t");
  Serial.print("HEX4"); 
  Serial.print("\t");
  Serial.print("HEX8"); 
  Serial.print("\t\t");
  Serial.print("OCT"); 
  Serial.print("\t");
  Serial.print("BIN");
  Serial.print("\t");
  Serial.print("BIN2");
  Serial.print("\t");
  Serial.print("BIN4");
  Serial.print("\t");
  Serial.print("BIN8");
  Serial.print("\t\t");
  Serial.print("BIN16");
  Serial.print("\t\t\t");
  Serial.print("BIN32");
  Serial.println(); // carriage return after the last label
  for (int x = 0; x < 64; x++)
  { // only part of the ASCII chart, change to suit
    // print it out in many formats:
    Serial.print(x); // print as an ASCII-encoded decimal - same as "DEC" 
    Serial.print("\t\t"); // prints two tabs to accomodate the label length
    Serial.print(x, DEC); // print as an ASCII-encoded decimal 
    Serial.print("\t"); // prints a tab
    Serial.print(x, HEX); // print as an ASCII-encoded hexadecimal 
    Serial.print("\t"); // prints a tab
    Serial.print(x, HEX2); // print as an ASCII-encoded hexadecimal 
    Serial.print("\t"); // prints a tab
    Serial.print(x, HEX4); // print as an ASCII-encoded hexadecimal 
    Serial.print("\t"); // prints a tab
    Serial.print(x, HEX8); // print as an ASCII-encoded hexadecimal 
    Serial.print("\t"); // prints a tab
    Serial.print(x, OCT); // print as an ASCII-encoded octal 
    Serial.print("\t"); // prints a tab
    Serial.print(x, BIN); // print as an ASCII-encoded binary 
    Serial.print("\t"); // prints a tab
    Serial.print(x, BIN2); // print as an ASCII-encoded binary 
    Serial.print("\t"); // prints a tab
    Serial.print(x, BIN4); // print as an ASCII-encoded binary 
    Serial.print("\t"); // prints a tab
    Serial.print(x, BIN8); // print as an ASCII-encoded binary 
    Serial.print("\t"); // prints a tab
    Serial.print(x, BIN16); // print as an ASCII-encoded binary 
    Serial.print("\t"); // prints a tab
    Serial.println(x, BIN32); // print as an ASCII-encoded binary 
    // then adds the carriage return with "println"
    delay(200); // delay 200 milliseconds
  }
  Serial.println(); // prints another carriage return
}
```

### DIFF
--- a/cores/arduino/Print.cpp
+++ b/cores/arduino/Print.cpp
@@ -69,41 +69,39 @@ size_t Print::print(char c)
   return write(c);
 }
 
-size_t Print::print(unsigned char b, int base)
+size_t Print::print(char num, int base)
 {
-  return print((unsigned long) b, base);
+  return printNumber ((long) num, num < 0, sizeof(char), base);
 }
 
-size_t Print::print(int n, int base)
+size_t Print::print(signed char num, int base)
 {
-  return print((long) n, base);
+  return printNumber ((long) num, num < 0, sizeof(char), base);
 }
 
-size_t Print::print(unsigned int n, int base)
+size_t Print::print(unsigned char num, int base)
 {
-  return print((unsigned long) n, base);
+  return printNumber ((unsigned long) num, false, sizeof(char), base);
 }
 
-size_t Print::print(long n, int base)
+size_t Print::print(int num, int base)
 {
-  if (base == 0) {
-    return write(n);
-  } else if (base == 10) {
-    if (n < 0) {
-      int t = print('-');
-      n = -n;
-      return printNumber(n, 10) + t;
-    }
-    return printNumber(n, 10);
-  } else {
-    return printNumber(n, base);
-  }
+  return printNumber ((long) num, num < 0, sizeof(int), base);
 }
 
-size_t Print::print(unsigned long n, int base)
+size_t Print::print(unsigned int num, int base)
 {
-  if (base == 0) return write(n);
-  else return printNumber(n, base);
+  return printNumber ((unsigned long) num, false, sizeof(int), base);
+}
+
+size_t Print::print(long num, int base)
+{
+  return printNumber (num, num < 0, sizeof(long), base);
+}
+
+size_t Print::print(unsigned long num, int base)
+{
+  return printNumber (num, false, sizeof(long), base);
 }
 
 size_t Print::print(double n, int digits)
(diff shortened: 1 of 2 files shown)


per1234 (commented):
Thanks for your pull request @TobiasKnauss!

Something important to understand is that the goal is for every Arduino boards platform to implement a standardized API, as described by the Arduino Language Reference:

https://docs.arduino.cc/language-reference/

Standardization is important for the sake of portability. If all cores implement a standardized API, then sketches and libraries that only use that API will work with every one of the diverse array of boards in the Arduino hardware ecosystem. Conversely, if we only implement a capability one core, then users will find that code utilizing that capability won't compile for boards of any other platform.

In order to facilitate standardization of cores, Arduino has created a reusable codebase that defines a standardized API, as well as provides the implementations of all components that are not architecture-specific:

https://github.com/arduino/ArduinoCore-API

You are proposing an expansion of the standardized core API. It is true that, if accepted, this should be implemented in the core of the "**Arduino AVR Boards**" platform hosted in this repository, but the same should also be done in the core of every other platform. For this reason, it is best to start by proposing the addition of the new capability in the `arduino/ArduinoCore-API` repository. If the maintainers of that project accept your proposal, then it will be clear that the introduction into the individual cores is appropriate.

So the procedure for proposing changes to the core API should be:

1. Submission of proposal to the `arduino/ArduinoCore-API` repository.
1. Review of proposal.
1. Adjustment of proposal as needed to resolve review.
1. Acceptance and implementation of proposal.
1. Propagate change to cores.

This pull request is needed to complete the final step in that procedure, so it is valuable and appreciated. However, the other steps in the procedure must be completed first.

TobiasKnauss:
Like in the API (awaiting confirmation of my PR there), I have also added further improvements here.

- Printing negative numbers in HEX always produced a 8-char-string, even if a byte (u/int8_t) or int (u/int16_t) was provided. The modifications now print in a length that matches the given variable type: byte -> 2-char HEX representation, int -> 4-char HEX representation, etc.

Test program and text output are added there: https://github.com/arduino/ArduinoCore-API/pull/271#issuecomment-4578101724

